PR #1, opened 2019-06-27
Mongo docker-compose example

wtfred:
My example of a MongoDB integration with ddev, please tell me if i need to improve anything :)


rfay:
Also remember that your contribution here will likely be the pattern for every service that will follow, so yay!

wtfred:
Ok, i'll try to work a bit more on the README next week :)

rfay:
BTW, it would probably be best to change the docker-composer for mongo express *not* to be accessible to the entire network. 
You do it like this: 
```
    ports:
      - "127.0.0.1:<hostport>:<containerport>"
```

wtfred:
Hello @rfay, i made the change you adviced, seems to works: i can access it, coworkers can't anymore.
Do i also need to change
```
ports:
      - "27017"
```
to 
```
ports:
      - "127.0.0.1:27017:27017"
```
for the `mongo` definition? Looks like it's not needed but i'm not fluent with ports and stuff so... :D

rfay:
Yes, I recommend changing to `127.0.0.1:27017:27017` - The difference with this one is that you were getting an ephemeral (un-reserved) port on localhost with 
```
ports:
      - "27017"
```

so people would have to work to find it, but it would be there waiting for them if they probed it. 

You should probably add a note that only one project at a time can run with this wired port, so a second project using this recipe will need altered ports.

Also, this doesn't work on docker toolbox, where the IP address is not 127.0.0.1, it's <docker-ip>, maybe 192.168.100.1. 

Thanks for all the work on this!

PR #4, opened 2019-07-10
Add hook example for post start import db if empty

schmitzal:


schmitzal:
Pull request authors can't approve their own pull request, so here comes my approval for your changes :D

rfay:
Thanks so much for this!

PR #5, opened 2019-08-02
Add blackfire example

susannemoog:


rfay:
I imagine you're already working on this, but please add a README.md with the full setup instructions, and an entry in the top-level README.md. Thanks!

susannemoog:
> 
> I imagine you're already working on this, but please add a README.md with the full setup instructions, and an entry in the top-level README.md. Thanks!

Added instructions that are hopefully understandable.

rfay:
You're awesome, I'll look forward to trying it out. 

I'm a never-ever with Blackfire. Should we/could we add at least a link to what you would do to try it out, how you'd know it was working?

susannemoog:
I'm unsure what would be needed. Basically after you followed the instructions (and added the client as described on the blackfire web page link) you should be able to start profiling requests either with the firefox or chrome addon or with the cli client as described here: https://blackfire.io/docs/cookbooks/profiling-http - does that help or do you need more info? if that's enough I can update the readme. if you need anything more, feel free to hit me up on TYPO3 slack.

rfay:
@mglaman @susannemoog I worked through this as a Blackfire n00b and made a number of updates in 8dc1fb2 which I'd appreciate if you'd review.

I'm not entirely convinced that we need the blackfire agent in a separate container, but whatever. Seems to me like we could just `webimage_extra_packages: [blackfire-agent]` and provide the config via environment variables as we're already doing. But it's fine like this I guess.

susannemoog:
@rfay something weird seems to have happened in the `basic usage` section - I guess you wanted to delete that? Additionally, if you are mentioning the chrome extension, should we also mention the one for firefox? (https://blackfire.io/docs/integrations/firefox )

And no, it's not strictly necessary to run the agent in a separate container, however it's cleaner and I think clearer that way, as in theory you'd need only one running agent for many projects connecting to it (and especially when thinking production setups and the like, making it clear that the agent is a separate service is nicer, I think).

rfay:
Something definitely happened, sorry! I'll fix.

rfay:
Fixed up basic usage in [readme](https://github.com/drud/ddev-contrib/blob/425b22da368d27f68bd53c26184a863ee32fe0b9/docker-compose-services/blackfire/README.md)

susannemoog:
Looks better now. What about the firefox addon? Or do you deem it not important enough? It's fine either way from my side.

rfay:
Sorry, I just forgot. Feel free to push a commit in there. Thanks for the review and the pr!

rfay:
Thanks so much for this!

PR #8, opened 2019-08-19
Added Elasticsearch service example, fixes drud/ddev#1659

dacostafilipe:
Example of the configuration we use internally for using Elasticsearch with ddev

In our case, we don't need to restore data after a stop/start.


rfay:
Sorry for the delay on reviewing this, it's very much appreciated.

rfay:
Thanks for updating!

rfay:
I took the liberty of rebasing to upstream.

PR #10, opened 2019-08-23
Drupal 8 multisite configuration recipe

arafalov:
As was discussed in /drud/ddev#1806, an end-to-end example to setup a Drupal 8 multisite within DDEV with explanations of extension points, files to override and support for Drush.

If accepted, it can be linked from the main document at the end of https://ddev.readthedocs.io/en/stable/users/cli-usage/#drupal-8-quickstart

### DIFF
--- a/recipes/drupal8-multisite/README.md
+++ b/recipes/drupal8-multisite/README.md
@@ -6,7 +6,7 @@ This example will show how to setup a basic two-site multisite configuration, co
 
 Neither of the sites will be default to ensure that if something is misconfigured, commands will fail against incomplete Drupal site instance instead of silently running against wrong one.
 
-## Initial setup
+## Initial DDEV-Local and Drupal 8 setup
 
 Let's mostly follow the suggested configuration in [the official DDEV documentation for Drupal 8 setup](https://ddev.readthedocs.io/en/stable/users/cli-usage/#drupal-8-quickstart).
 
@@ -25,7 +25,7 @@ Let's create two sites:
 * **basic** - using *standard* install profile
 * **umami** - using *umami* install profile
 
-## Preparing the databases and URLs
+## Prepare the databases and URLs
 
 The file *.ddev/config.yaml* is where additional URLs are setup and where we can use hooks to create the databases. However, it may be better to use [a separate override config file](https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/#extending-configyaml-with-custom-configyaml-files) to keep that information more clear and easier to track. The combination of sites may also be different on each individual machine, so this file may or may not be checked into the version control.
 
@@ -40,14 +40,14 @@ You should get a message that additional domains are now available:
 
 > Your project can be reached at <https://basic.ddev.site> <https://d8m.ddev.site> <https://umami.ddev.site> <https://127.0.0.1:32817>
 
-## Preparing drush
+## Enable the `ddev drush` custom command
 
-Drush is useful, but it is a bit annoying to *ddev ssh* into the container all the time or to run *ddev exec drush*. Fortunately, latest version of DDEV already ships with a simpler way, using [Container commands](https://ddev.readthedocs.io/en/stable/users/extend/custom-commands/).  And drush is one of the examples, so it just needs to be renamed.
+Drush is useful, but it is a bit annoying to *ddev ssh* into the container all the time or to run *ddev exec drush*. Fortunately, DDEV has [Custom commands](https://ddev.readthedocs.io/en/stable/users/extend/custom-commands/).  And drush is one of the examples, so it just needs to be renamed.
 
 1. mv .ddev/commands/web/drush.example .ddev/commands/web/drush
 2. ddev drush status
 
-## Enabling multisite
+## Enable multisite
 
 First we need to enable the multisite support by copying *example.sites.php* to *sites.php*. And then, because we are using DDEV for development and our production URLs will be different from test URLs, we want to define explicit aliases. That also allows us to have nice site directory names.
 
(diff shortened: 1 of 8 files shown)


rfay:
Yay!

rfay:
Looking forward to minor updates to get this in @arafalov , and thanks so much for the comment @Gonzalo2683

recipes/drupal8-multisite/README.md
```diff
@@ -0,0 +1,152 @@
...
+Neither of the sites will be default to ensure that if something is misconfigured, commands will fail against incomplete Drupal site instance instead of silently running against wrong one.
+
+# Initial setup
+
+Let's mostly follow the suggested configuration in [the official DDEV documentation for Drupal 8 setup](https://ddev.readthedocs.io/en/stable/users/cli-usage/#drupal-8-quickstart).
+1. mkdir d8m
+1. cd d8m
+1. ddev config --project-type php
+1. ddev composer create drupal-composer/drupal-project:8.x-dev --stability dev --no-interaction --prefer-dist
+1. ddev config --project-type drupal8
```
wizonesolutions: This guide looks amazing. I'm wondering if the `php` project type would be a better choice, though. The `drupal8` project type tries to re-add `settings.ddev.php` to `sites/default/settings.php` every time. This might be solved by not _having_ a `default` site, though, thus a non-issue? `drupal8` _is_ probably a better project type because I have found that when using `ddev config` to upgrade my configuration, ddev re-detects the project as `drupal8` anyway.

I am planning to set up a multisite config soon, so I will keep this recipe in mind and see if I can contribute anything.

rfay:
This is super important, and I really look forward to pulling it.  The one thing that is a real blocker for me is drush not really getting the right URL yet, would love to solve that and get this in. Thanks for it!

arafalov:
Sorry, I had a combination of external deadlines, personal leave and being sick. On the other hand, I tested the workflow from here on internal system. 

Things are easier now, I am hoping to incorporate the comments and do the updated submission this week.

rfay:
Looking forward to getting this in. Mostly it's fantastic, just that question about drush urls

Graloth:
From my testing this seems to work without much issue, though there are some things that I would prefer to be handled differently (this is mostly personal preference, mentioning it here in case it makes enough sense to be included).

- The ddev command `ddev import-db` doesn't seem to support a parameter to specify what database to import into (or at least it's --help doesn't list it as having the option)
- Similarly, the `ddev sequelpro` doesn't support specifying a database to open, it defaults to `db` which is always empty in our case. Sequel Pro does however, once opened, have access to navigate to the other database(s)

Also, the setup we usually run is that we use a multisite setup even though the website is only one site, meaning we never use the `default` site folder.
This creates a minor issue when configuring `config.multisite.yaml` as the hostname for the site mentioned in it is the same as the name mentioned in `config.yaml` and as such the ddev router creates two sites with the same hostname (and thus it appears twice when using `ddev list`)

Example:

- The `config.yaml` name is set to "testsite", so we get the fqdn `testsite.ddev.site`
- The `config.multisite.yaml` has one entry for "testsite" as an additional hostname, which also get s the fqdn `testsite.ddev.site`.

This is of course intended behavior, as we're defining an additional hostname that is identical to the ddev project name and thus the router handles both. (and the site still works fine)

I am mostly mentioning this as the documentation could benefit from explaining that if you use multisite with only one site, change the name to match the hostname you want and don't add additional hostnames inside the multisite config. Or simply in that case ignore the `config.multisite.yaml` and use the normal database as well as change the name to match the hostname you want.

Another suggestion might be to have the router not add a route that already exists, or at least that `ddev list` doesn't list identical routes to keep the list concise.

---

Apart from what is mentioned above, this documentation works for me, but it would be great if we could define which sites ddev should create it's `settings.ddev.php` inside, so that we can have it autogenerate it only in the sites we want (though I only consider this a quality-of-life addition)

This documentation is definitely one that will benefit others, so I hope it'll get added to the docs soon :)

rfay:
Thanks for the review @Graloth ! 

Just a note : `ddev mysql` is often more useful for extensive commands than import-db. It has everything you'd ever want (because it just runs mysql in the db container). So it's probably a better way to create and populate extra databases.  However, https://github.com/drud/ddev/issues/1652 does request adding a db flag to import-db. 

I didn't understand 
> Another suggestion might be to have the router not add a route that already exists, or at least that ddev list doesn't list identical routes to keep the list concise.

If the router is showing two identical projects (same name) something is really wrong. How does that happen?

Graloth:
> If the router is showing two identical projects (same name) something is really wrong. How does that happen?

In my case it happened after following this guide, where in my `config.yaml` "name" was set to the same as the additional hostname I added in `config.multisite.yaml`, which made it appear twice when using `ddev list`, not sure if it means that two routes were actually created, now that I think more about it it's probably just listed twice because it was defined twice (once for name and another for the additional hostname).
Example:

- Default `config.yaml`'s name variable set to "testsite"
- Add two additional hostnames to `config.multisite.yaml`, "testsite" and "testsite2"
- Start the project
- Run `ddev list` and see that you have the "testsite" showing twice (one from the name variable and one from the additional hostname) and "testsite2" showing only once (since it's only defined as an additional hostname)

I have also just noticed that the default gitignore for the .ddev folder, ignores all config.*.yaml, meaning that the multisite config is ignored. This might not be optimal as I can imagine this configuration being one you'd definitely want to share with other devs on the project. Naturally we can manually add the file to git to circumvent this, in which case that might be worth mentioning in the docs.

rfay:
Interesting, now why would you add "testsite" additional hostname to a project called "testsite"? But it's a bug if that shows up twice then in `ddev list`. Or where does it show up twice? If it's a bug, would appreciate an issue in ddev queue.

Graloth:
> Interesting, now why would you add "testsite" additional hostname to a project called "testsite"? But it's a bug if that shows up twice then in `ddev list`. Or where does it show up twice? If it's a bug, would appreciate an issue in ddev queue.

I've only seen it twice in `ddev list` and I noticed it after I was running on autopilot and added the additional hostname without realising I wasn't needed :) so I don't consider it a bug really, but might be worth a short note somewhere that you don't need to add an additional hostname if the name of the project is the same... at least for idiots like me :P

I'll see if I can replicate it from scratch again and if so make an issue like you requested

mac-dev:
We recently updated ddev to 1.12 from 1.8.0 and this broke our D7 Multisite install.
There are autgenerated drushrc.php and settings.php files which cause the site always go to the default database db even while drush is called from a sites subdir. 
We could delete those files in our setup routine written in python which does similar things like in the toutorial above but a bit more as we import DBs from production and so on. Maybe it would be usfull to share this script.
Anyway - is there a flag to stop the autcreation of the settings and drushrc files in sites/default ?

wizonesolutions:
Use type php instead of drupal7

rfay:
@mac-dev @wizonesolutions is correct: If you use project type "php" , which is outstanding for advanced users, ddev won't create any settings files or the like.

rfay:
Rebased and made minor edits trying to prepare to finally get this in!

rfay:
Yay, thanks all!

PR #12, opened 2019-09-04
add mysql backup and restore commands

thomaskieslich:
In addition to the docker-compose.mysql.yaml, if you like ;-)


rfay:
Thanks for this! Please do sign the CLA.

I think we probably want a new top-level section "Commands" for commands, there should be lots of them like this.

recipes/mysql-backup-restore/commands/mysql/mysql-backup
```diff
@@ -0,0 +1,18 @@
...
+## Description: Backup MYSQL DB quick or with snapshot
+## Usage: mysql-backup
+## Example: "ddev mysql-backup snapshot"
+## Options: "mysql-backup snapshot - save backup with timestamp"
+
+# create dir if not available
+mkdir -p /mnt/ddev_config/backup
+
+if [[ $1 = "snapshot" ]]; then
+    mysqldump --single-transaction --quick --lock-tables=false -udb -pdb db | gzip > /mnt/ddev_config/backup/mysql-db-$(date +"%Y%m%d-%H%M").sql.gz
```
rfay: In my own workflow I'd rather just have mysqldump output to stdout, where I can do with it as I will.

rfay:
I do think this should just be added into the mysql recipe, https://github.com/drud/ddev-contrib/tree/master/docker-compose-services/mysql, I think that's what you were suggesting as well.

recipes/mysql-backup-restore/commands/mysql/mysql-backup
thomaskieslich: Feel free to edit the commands but we need this commands for quick and flexible Backup/Restores.

thomaskieslich:
updated and moved to docker-compose-services/mysql

rfay:
Thanks so much for this!

PR #14, opened 2019-09-15
Fix link in blackfire example, fixes #13

### LINKED ISSUE #13: blackfire: Service conflict when running multiple projects at the same time
When you have multiple ddev projects running, and you're using blackfire in all of them, then you will most likely talk to the wrong service when connecting to `tcp://blackfire:8707`.

see
- https://github.com/drud/ddev-contrib/blob/aeacfb338cf3da2d7dee61e4741b2aeca70c5c95/docker-compose-services/blackfire/docker-compose.blackfire.yaml#L13
- https://github.com/drud/ddev-contrib/blob/aeacfb338cf3da2d7dee61e4741b2aeca70c5c95/docker-compose-services/blackfire/php/blackfire.ini#L1

The service is defined like this:

```
services:
  web:
    links:
      blackfire:$DDEV_HOSTNAME
```

This means that the container `blackfire` is aliased as `$DDEV_HOSTNAME`. This doesn't work, and I assume it is silently ignored by Docker...
The problem is, when you still connect to the host `blackfire`, then this name exists multiple times. Try to ping the `blackfire` host from the `web` container multiple times to see this in effect...

Therefore, connecting to `blackfire.agent_socket = tcp://blackfire:8707` will only work if the host is given a distinct alias.

This can be solved in a very simple way: Just assign the alias `blackfire` to this container, and it will work as expected:

```
services:
  web:
    links:
      blackfire:blackfire
```

When you ping the `blackfire` host again, it will always resolve to the service from the same project.

This is also solved in the same way for mongodb in this project, see
https://github.com/drud/ddev-contrib/blob/aeacfb338cf3da2d7dee61e4741b2aeca70c5c95/docker-compose-services/mongodb/docker-compose.mongo.yaml#L43

stucki:
Closes #13

### DIFF
--- a/docker-compose-services/blackfire/docker-compose.blackfire.yaml
+++ b/docker-compose-services/blackfire/docker-compose.blackfire.yaml
@@ -10,4 +10,4 @@ services:
     hostname: ${DDEV_SITENAME}-blackfire
   web:
     links:
-      - blackfire:$DDEV_HOSTNAME
+      - blackfire:blackfire


rfay:
Thanks so much for this!

PR #17, opened 2019-10-01
Create fetch-production-db

kaystrobach:
example for fetch the production db from the remote server with ddev


rfay:
Thanks!

hook-examples/fetchproductiondb/fetchproductiondb
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+## Description: Open PHPStorm with the current project
+## Usage: fetchproductiondb
+## Example: "ddev fetchprodctiondb"
+
+# Example is macOS-specific, but easy to adapt to any OS
+ssh -t user@host 'mysqldump --defaults-file=~/.mysql/project.cnf databasename' > productionsqldump.sql
+ddev import-db --src=productionsqldump.sql
```
rfay: FYI `ddev import-db` (and also `ddev mysql`) are happy to accept from stdin, so it's not actually necessary to dump to a file first. (Or end up having to delete the file)

kaystrobach: mhm ok, we decided to have the step as we can then easier debug the output by simply adding a pause or commenting the rm - but i will think about that

kaystrobach: the web command sounds like a good plan - will check it out

rfay:
Another suggestion might be to use "pullprod" or something as the verb here, matching `ddev pull`

rfay:
Rebased, Moved to custom command, made several edits, moved config into variables. Hope you like it! Thanks for this!

PR #18, opened 2019-10-03
Added a Laravel recipe, fixes drud/ddev#898

tommym9:


rfay:
You're awesome, thanks!

rfay:
Thanks for reviewing @tomasnorre !

tomasnorre:
You're welcome..

rfay:
I 
* rebased this to current master
* Added a link in the top-level README
* Changed to use `ddev composer create` in the composer instructions

It worked perfectly... but I was confused by the "what next"; I imagine that I'm not the only one.

I recommend *only* providing the composer instructions, and then the artisan stuff can be removed?

And could we add a "What do I do now with that beautiful blank Laravel screen?

Screenshot_10_12_19__4_10_PM

tomasnorre:
You start developing.. :) 

It's just a default like "Apache/Nginx" startpage to prove that the setup is working. 

If you see that screen it's working.

PR #26, opened 2019-10-21
Improve custom PHP instructions for blackfire, fixes #25

### LINKED ISSUE #25: Clarify documentation for Blackfire
related to https://github.com/drud/ddev/issues/1606

It is not obvious that the blackfire.agent_socket needs to be overriden in order to work. It's also not clear on how this gets overridden. The current README hints that the agent socket can be adjusted via the included blackfire.ini, which can be misinterpreted that since it's included, it's being automatically pulled in.

I think we should clarify that the user will need to explicitly add a custom php.ini configuration file in order to get blackfire working correctly.

briantully:
In trying to get blackfire working, it wasn't obvious to me that the blackfire.agent_socket needed to get overridden and added to a custom php.ini file as outlined by the DDEV docs. Because of the way the current blackfire README was worded, I assumed that the php/blackfire.ini would automatically get pulled in.

This PR updates the README to try to make it more obvious that there is a manual step required, and links to the DDEV docs on how to create a custom php.ini.

### DIFF
--- a/docker-compose-services/blackfire/README.md
+++ b/docker-compose-services/blackfire/README.md
@@ -14,10 +14,8 @@ The probe is a PHP module which is already installed on a default DDEV setup.
 
 The agent is provided as docker image by blackfire. Add it to your setup via the provided [docker-compose.blackfire.yaml](docker-compose.blackfire.yaml). Make sure to edit the `BLACKFIRE_SERVER_ID` and `BLACKFIRE_SERVER_TOKEN` environment variables set there.
 
-To make sure the probe and agent can communicate, you will need to override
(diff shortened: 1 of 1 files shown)


rfay:
Thanks so much for this. I'm quite sure I didn't have this problem when testing the PR that introduced this config, but I'll test again, and maybe we can get to the root of the issue that caused you trouble.

rfay:
I made a couple of minor edits @briantully - could you take  a look at https://github.com/drud/ddev-contrib/blob/2df4dd095e0ff444533fa6f094d0153ec785ac52/docker-compose-services/blackfire/README.md and see if it does the job?

briantully:
Awesome! It definitely does the job. Much appreciated :)

rfay:
Thanks so much for taking the initiative on this!

PR #29, opened 2019-11-05
RabbitMQ service example.

Graloth:
Adds a RabbitMQ service example with readme explaining the basics of how to connect to it.

We've tested this in our local ddev environments and it works fine, though if you need to run multiple ddev sites with this service, you will have to alter the ports so they do not interfere with each-other.


rfay:
Sorry haven't reviewed yet, very much appreciate the PR, it will get going before too terribly much longer.

rfay:
Made minor edits, added link in top-level readme. Thanks!

PR #30, opened 2019-11-18
add custom commands dump-db/deploy-db

jonaseberle:
This adds custom commands to facilitate working with remote databases


rfay:
Awesome, thanks!  Not sure why it still wants you to sign CLA, but please do it anyway. Do you use more than one email address?

jonaseberle:
> Do you use more than one email address?

Yes. It confuses myself, too. I tried to use 1 for work and 1 for private but I am mixing it up...

rfay:
I noted that the "target" directory showed up as a command, fixed that in https://github.com/drud/ddev/pull/2022

rfay:
I do find this incomprehensibly slow in both directions. I guess if you want it that way, you want it that way! Maybe we need more explanation of why you'd want to wait for many minutes to do something that normally takes a few seconds. I know this has explicit value.

rfay:
Rebased and made minor edits. Tried out a target to dump from. It seems to be working, but not too tolerable from a time perspective.

rfay:
Thanks so much!

jonaseberle:
It opens the SSH tunnel only once but the MySQL connection is initiated for each SQL file. This is to be able to manage SQL files by struct/table.

PR #34, opened 2019-12-27
Add FlextypeCMS recipe

crydotsnake:
If you guys have any ideas what i could change, just let me know!.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,8 @@ General information on how to do additional services and some additional example
 
 * [Using ddev with a corporate (or other) web proxy](recipes/proxy/README.md)
 * [enable cronjob on start or on demand (typo3 scheduler:run)](recipes/cronjob/README.md)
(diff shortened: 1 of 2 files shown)


rfay:
I made a couple of minor wording changes (so you'll have to `git pull`

But work remains. Please:

* Explain how you get the code for flextype (you left that part out).  
* Add a link in the top-level README.md

There's no actual difference between this recipe and the generic PHP recipe in https://ddev.readthedocs.io/en/stable/users/cli-usage/#quickstart-guides - it's useful to show here that flextype works with ddev, so that's great, but a technique like this should also show anything unusual about how it's used.

crydotsnake:
Maybe it doesent make sense to write i recipe for installing a flat file cms with ddev. But i have no recipe ideas.

rfay:
Thanks for this!

PR #35, opened 2020-01-06
Document blackfire.io cli client setup and usage

maddy2101:


rfay:
Thanks so much for exploring this territory and taking the time to document this!

PR #37, opened 2020-02-10
Update elasticsearch service example

Graloth:
This is just a simple update, as it seems to me that the common naming scheme for additional services is that they are separated by a dot (.) not a dash (-), so this pull request just fixes that and a minor typo in the readme.

No functional changes were made to the service itself.


dacostafilipe:
Yes, this seems right.

PR #38, opened 2020-02-10
Add elastichq service example

Graloth:
Adds an elastichq service example that integrates with an existing elasticsearch service within the DDEV project, allowing you to have a GUI for the contents of your elasticsearch server.

I made this service as I needed a proper way to confirm that my data was properly stored on the elasticsearch server without having to run a bunch of terminal commands.


rfay:
I'm going to wait just a bit for the other one to land, and then we'll get this one in there. Should they be in one section?

PR #36, opened 2020-01-15
Add Redis service and CLI command

gormus:

PR #39, opened 2020-02-10
Adds a redis service example

Graloth:
This will allow your web container to use redis.

It still needs a readme, which I haven't gotten around to writing, as it can be helpful for people to know how they actually are able to connect to the redis server (for example, how to do it in Drupal).

### DIFF
--- a/docker-compose-services/redis/docker-compose.redis.yaml
+++ b/docker-compose-services/redis/docker-compose.redis.yaml
@@ -1,0 +1,20 @@
+version: '3.6'
+services:
+  redis:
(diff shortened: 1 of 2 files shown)


Graloth:
I've also added a service for redis commander, for those that want to be able to view the data on the redis server.

Graloth:
I just noticed Pull Request #36, so I'll close this pull request and work on updating rediscommander to work with that instead as it has both documentation and a CLI command shortcut.

rfay:
Thanks @Graloth - Could you review #36 please? Thanks!

PR #41, opened 2020-02-11
Adds an example of a redis-commander service.

Graloth:
This is a service example with a readme for adding a GUI client to your redis server.

I've added a proper readme to this example, as I have more experience with it.

This service is tested to work with the redis service example from PR #36 which is also my recommended way to run redis.


rfay:
I'll wait until the #36 lands so we can properly link them, but should be great. Thanks!

rfay:
Worked great for me. I added https capability and made minor edits to readme, etc.

PR #42, opened 2020-02-12
Direct to correct drush.yaml file

amitaibu:
- [ ] multisite PR maintenance
- [ ] Use disable_settings_management: true
- [ ] Remove reference to drush' @self.site.yml
- [ ] Better example of site settings.php file
- [ ] drush.yml explanation of emptying it is in the wrong place
- [ ] root: in the files in drush/sites is all wrong (has root: /var/www/html/web/basic, should be root: /var/www/html/web/sites/basic)

### DIFF
--- a/recipes/drupal8-multisite/README.md
+++ b/recipes/drupal8-multisite/README.md
@@ -65,7 +65,7 @@ Now we have to do a DDEV specific thing. Trying to be helpful, it creates a glob
 
 We want to disable that option. However, the file is DDEV generated, so we can't just comment the line out, it may get regenerated. We need to remove the auto-generation marker lines as well. We can make this file empty.
 
-1. Edit `web/sites/all/drush/drush.yml` and delete all content including comments
+1. Edit `drush/drush.yml` and delete all content including comments
 2. Explicitly check the file in to git, if you use git (it has to be explicit, as DDEV also creates a local `.gitignore` file)
 
 ## Prepare example umami site


amitaibu:
Seems that maybe some further changes are needed (which I haven't figured out myself yet), around https://github.com/drud/ddev-contrib/blob/master/recipes/drupal8-multisite/README.md#set-up-drush-site-aliases

image

Currently the docs have the alias as for example `@umami.ddev`, however currently I'm seeing differently:

image

rfay:
Thanks @amitaibu - Complex recipes like this always need maintenance, and your work here is much appreciated.  It looks to me like it should be `ddev drush @umami` instead of `@umami.ddev` ?

amitaibu:
> and your work here is much appreciated

Are you kidding? It's *your* (et al) that allows me to spin it so quickly. So :medal_military: for you folks :)

> It looks to me like it should be ddev drush @umami instead of @umami.ddev

Indeed, however I get only `umami` (without `@`). So currently drush isn't working for me. Here's an [example](https://github.com/amitaibu/multi-repo/tree/4980e631537074847eab515183875dd3a5c84a95) repo I've been trying with -

rfay:
Oh sorry, it should definitely be `ddev exec drush ...` if you haven't turned on the optional "drush" example command

rfay:
I recommend (in general) using `ddev ssh` and then use drush there when wanting to do stuff in the container. It's faster than `ddev exec drush`

amitaibu:
> ddev exec drush

I have turned it on. But also with it, I get only `umami`

image

image

rfay:
I've never seen a repo layout like the one you show. Where's the composer.json? Have you done `composer require drush/drush` ? Otherwise you're stuck with drush8 as the fallback inside the container. 

I recommend sticking closer to the classic drupal-composer/drupal-project or drupal/recommended-project...

amitaibu:
Good catch. I have missed the `require drush/drush` as the multi-site docs link to the stable ones, instead of latest.

image

I'll re-try to scaffold.

amitaibu:
btw, on the latest docs, it says `ddev start` - shouldn't it say `ddev restart`?

image

amitaibu:
@rfay I've created a [new scaffold](https://github.com/amitaibu/multi-repo/tree/ed09f474883220173c3e750e1280aea732f06f16) - hopefully I got it right this time :smile_cat: 

However, now I don't seem to get any alias.

image

PR #48, opened 2020-03-16
add volume to postgres config to allow persistence

d3pendent:


rfay:
Everybody seems to like it, thanks @d3pendent - I removed the line in readme that implies this is a future enhancement @gilbertsoft .

PR #51, opened 2020-03-28
Adds an example of selenium service

tomasnorre:
This is a small example on how selenium can be added to ddev to e.g. run Codeception tests locally.

### DIFF
--- a/docker-compose-services/selenium/README.md
+++ b/docker-compose-services/selenium/README.md
@@ -1,0 +1,15 @@
+# Selenium in DDEV  
+
+Running selenium for e.g Codeception tests can be quite handy.
+
+To include this to your own project you can do following:
+
+```
+$ cp docker-compose.selenium.yaml /path/to/your/project/.ddev/docker-composer.selenium.yaml
+$ cd /path/to/your/project/
+$ ddev start # or ddev restart (depending on the state of the project)
+```
+
+Now you have a selenium with a chrome headless browser node running in your ddev project.
+
+To see how to use it with Codeception, you can take a loot at (https://dev.to/tomasnorre/codeception-ddev-selenium-docker-36kk)[https://dev.to/tomasnorre/codeception-ddev-selenium-docker-36kk]
--- a/docker-compose-services/selenium/README.md
+++ b/docker-compose-services/selenium/README.md
@@ -12,4 +12,6 @@ $ ddev start # or ddev restart (depending on the state of the project)
 
 Now you have a selenium with a chrome headless browser node running in your ddev project.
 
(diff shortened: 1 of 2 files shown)


rfay:
Yay!

tomasnorre:
Yeah.. Will have a look.

tomasnorre:
Looks like the "same", but the difference is that the example you're linking to is using the Chrome stand-alone only, and not the Selenium as well.

I would need to check up on that if I did "to much".

I'll add some information on how this container itself can be tested. 
Thanks for your feedback.

rfay:
Ping @tomasnorre , looking forward to getting this in.

tomasnorre:
Pong ;) Yeah, me too. Got away from it as I was focusing on the release of the TYPO3 Crawler. 

Will see when I'll get to this, I honestly forgot about it because of my focus somewhere else.

tomasnorre:
As I'm not really used to use selenium, I'm not sure what else to add, if you have any ideas, please let me know.

tomasnorre:
BTW. Do I need to sign CLA? and if so how do I do that?

rfay:
You have already signed the CLA, thanks! It doesn't have to be done for every PR.

tomasnorre:
Ok. I wasn't aware of that. But good to know :D 

I knew i didn't have to every time, but didn't know I already did :)

Michael-Brooks:
@tomasnorre I've used the files from this PR to run Laravel Dusk tests and it works really well, thank you. However, I did get an error regarding volumes. I changed line 13

`- /dev/shm:/dev/shm`
to
`- /ddev/shm:/ddev/shm`

Which fixed my volume error. Not sure if this is an issue other people may come across.

tomasnorre:
Happy that I was useful for you too. 

The problem you are referring to is unknown to me. I will test if your suggestion works in my setup too if so I have no problem using your suggestion.

rfay:
@Michael-Brooks /dev/shm is a special linux driver for shared memory. You don't mention what kind of error you got, but it wouldn't make any sense to change "dev" to "ddev" there.

I don't understand why one would make /dev/shm a volume, so that line doesn't make any sense to begin with (to me). 

(Looking at @tomasnorre 's source, https://github.com/SeleniumHQ/docker-selenium#version-3, it looks like this volume was possibly put there to allow two containers to communicate with each other)

rfay:
@tomasnorre I apologize, but I haven't gotten you to add anything here that would explain how to manually test and validate this, despite several requests starting March 30. It seems like a really valuable PR, and it will be excellent, but it needs much more detail, especially manual testing detail, and explaining that it's not just selenium but chromedriver. 

I'm going to close this for now, but let's get it reopened and get it merged!

tomasnorre:
It's perfectly fine. I have so many open things. I'll keep it on my todo list and when I get to it, I'll reopen.

PR #54, opened 2020-04-06
Improve postgres example by mentioning client installation

rvolk:
The standard ddev-webserver comes with PHP pgsql extension enabled, but lacks the psql client. This little fix to your config file will install the  extra webimage package for postgresql-11. This is especially important when you run Drush for a Drupal site on a PostgreSQL database, which will fail if the psql command isn't available.

### DIFF
--- a/docker-compose-services/postgres/README.md
+++ b/docker-compose-services/postgres/README.md
@@ -39,7 +39,7 @@ Example `config.yaml` hooks configuration to automatically import/export the `db
 
 ```
 # Add psql to your webserver
-webimage_extra_packages: [postgresql-11]
+webimage_extra_packages: [postgresql-client]
 
 # Automatically import and export PostreSQL database on ddev start and stop
 hooks:
--- a/docker-compose-services/postgres/README.md
+++ b/docker-compose-services/postgres/README.md
@@ -38,6 +38,10 @@ Two new `ddev` commands are provided:
 Example `config.yaml` hooks configuration to automatically import/export the `db` table:
 
 ```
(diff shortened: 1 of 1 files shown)


docker-compose-services/postgres/README.md
```diff
@@ -38,6 +38,10 @@ Two new `ddev` commands are provided:
 Example `config.yaml` hooks configuration to automatically import/export the `db` table:
 
 ```
+# Add psql to your webserver
+webimage_extra_packages: [postgresql-11]
```
rfay, line 42: This should be postgresql-client, not postgresql-11, which would install the whole package, including server.

rvolk: Right, less is more 👍

docker-compose-services/postgres/README.md
```diff
@@ -38,6 +38,10 @@ Two new `ddev` commands are provided:
 Example `config.yaml` hooks configuration to automatically import/export the `db` table:
 
 ```
+# Add psql to your webserver
+webimage_extra_packages: [postgresql-client]
+
+# Automatically import and export PostreSQL database on ddev start and stop
```
rfay, line 44: Just an FYI, the hooks to import and export aren't needed any more as it's stored in a volume in the current version.

rvolk:
> This looks good. We should probably remove the export/import hooks as they're not needed any more. You can do it or I'll add a commit.
This relates to the persistence change in #48. I'll make a second request to keep this one focus.

PR #55, opened 2020-04-07
Add SSL info to Elasticsearch

amitaibu:
And enable CORS

### DIFF
--- a/docker-compose-services/elasticsearch/README.md
+++ b/docker-compose-services/elasticsearch/README.md
@@ -8,7 +8,7 @@ Copy [docker-compose.elasticsearch.yaml](docker-compose.elasticsearch.yaml) to y
 
 ## Configuration
 
-From within the container, the elasticsearch container is reached at hostname: elasticsearch, port: 9200, so the server URL might be `http://elasticsearch:9200`
+From within the container, the elasticsearch container is reached at hostname: elasticsearch, port: 9200, so the server URL might be `http://elasticsearch:9200`. You can also use the non-SSL, and SSL urls to access it: `http://<DDEV_SITENAME>.ddev.site:9200`, and `https://<DDEV_SITENAME>.ddev.site:9201`
 
 ## Connection
 
--- a/docker-compose-services/elasticsearch/README.md
+++ b/docker-compose-services/elasticsearch/README.md
@@ -8,11 +8,11 @@ Copy [docker-compose.elasticsearch.yaml](docker-compose.elasticsearch.yaml) to y
 
 ## Configuration
 
-From within the container, the elasticsearch container is reached at hostname: elasticsearch, port: 9200, so the server URL might be `http://elasticsearch:9200`.
+From within the container, the elasticsearch container is reached at hostname: elasticsearch, port: 9200, so the server URL might be `http://elasticsearch:9200`
 
 ## Connection
 
-You can access the Elasticsearch server directly from the host for debugging purposes by visiting `http://<DDEV_SITENAME>.ddev.site:9200`
+You can access the Elasticsearch server directly from the host for debugging purposes by visiting `http://<DDEV_SITENAME>.ddev.site:9200`. If you have SSL enabled, which is recommended, you can access Elasticsearch via `https://<DDEV_SITENAME>.ddev.site:9201`
 
 ## Memory Limit
 
--- a/docker-compose-services/elasticsearch/docker-compose.elasticsearch.yaml
+++ b/docker-compose-services/elasticsearch/docker-compose.elasticsearch.yaml
@@ -15,12 +15,6 @@ services:
       - VIRTUAL_HOST=$DDEV_HOSTNAME
       - HTTP_EXPOSE=9200:9200
       - HTTPS_EXPOSE=9201:9200
-
-      # Allow CORS.
-      - http.cors.enabled=true
-      - http.cors.allow-origin=*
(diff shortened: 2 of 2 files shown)


docker-compose-services/elasticsearch/README.md
```diff
@@ -8,11 +8,11 @@ Copy [docker-compose.elasticsearch.yaml](docker-compose.elasticsearch.yaml) to y
 
 ## Configuration
 
-From within the container, the elasticsearch container is reached at hostname: elasticsearch, port: 9200, so the server URL might be `http://elasticsearch:9200`.
+From within the container, the elasticsearch container is reached at hostname: elasticsearch, port: 9200, so the server URL might be `http://elasticsearch:9200`
 
 ## Connection
 
-You can access the Elasticsearch server directly from the host for debugging purposes by visiting `http://<DDEV_SITENAME>.ddev.site:9200`
+You can access the Elasticsearch server directly from the host for debugging purposes by visiting `http://<DDEV_SITENAME>.ddev.site:9200`. If you have SSL enabled, which is recommended, you can access Elasticsearch via `https://<DDEV_SITENAME>.ddev.site:9201`
```
rfay, line 15: You can also access it from both of those URLs inside the web container. So if people think the webserver should talk to elasticsearch via https they can use this URL. Probably a hair slower than using `http://elasticsearch: 9200`

docker-compose-services/elasticsearch/docker-compose.elasticsearch.yaml
```diff
@@ -15,6 +15,12 @@ services:
       - VIRTUAL_HOST=$DDEV_HOSTNAME
       - HTTP_EXPOSE=9200:9200
       - HTTPS_EXPOSE=9201:9200
+
+      # Allow CORS.
+      - http.cors.enabled=true
```
rfay, line 20: These are really lower-case environment variables named like `http.cors.enabled` ???
And what does this addition have to do with the stated goal of adding SSL info?
And what process consumes these environment variables?

docker-compose-services/elasticsearch/README.md
amitaibu: > You can also access it from both of those URLs inside the web container

I tried that, but it didn't work for me.

docker-compose-services/elasticsearch/docker-compose.elasticsearch.yaml
amitaibu: Right, might have needed a different PR - sorry. 

My use case is that I'm calling ES directly from client side via JS. That didn't work, due to CORS errors until I've added those environment variables?

docker-compose-services/elasticsearch/README.md
amitaibu: image

docker-compose-services/elasticsearch/docker-compose.elasticsearch.yaml
rfay: Client side where? Those variables get injected into the web container... so if they're going to affect something it's going to be nginx? Anyway, unless we know for sure why we're adding them, we can't have them in the PR.

docker-compose-services/elasticsearch/README.md
rfay: I didn't say you could use https://elasticsearch. I said you could use https://<projectname>.ddev.site:9201 - that goes through the router. The elasticsearch container you're using does not have mkcert respect or https built into it.

docker-compose-services/elasticsearch/docker-compose.elasticsearch.yaml
amitaibu: My use case is that my client side app was calling ES directly, so I needed to enable CORS

image

I found that config [here](https://github.com/appbaseio/dejavu/issues/256) -- and it worked for me.

So I'll remove this from this PR, question is if you'd want it in another PR -- maybe have that code commented out, just to save others the time to find this?

docker-compose-services/elasticsearch/README.md
```diff
@@ -8,11 +8,11 @@ Copy [docker-compose.elasticsearch.yaml](docker-compose.elasticsearch.yaml) to y
 
 ## Configuration
 
-From within the container, the elasticsearch container is reached at hostname: elasticsearch, port: 9200, so the server URL might be `http://elasticsearch:9200`.
+From within the container, the elasticsearch container is reached at hostname: elasticsearch, port: 9200, so the server URL might be `http://elasticsearch:9200`
```
amitaibu, line 11: @rfay is this what you mean?

```suggestion
From within the container, the elasticsearch container is reached at hostname: elasticsearch, port: 9200, so the server URL might be `http://elasticsearch:9200`. You can also use the non-SSL, and SSL urls to access it: `http://<DDEV_SITENAME>.ddev.site:9200`, and `https://<DDEV_SITENAME>.ddev.site:9201`
```

docker-compose-services/elasticsearch/docker-compose.elasticsearch.yaml
rfay: Makes sense, but we need to know why. And yes, let's get it into a different PR. I guess those must change the behavior of the java webserver elasticsearch is using. It would probably be better to incorporate them into launch args for elasticsearch, but I'm OK with this approach if we know why we're doing it. Must be in the elasticsearch docs.

PR #63, opened 2020-05-06
Update blackfire agent use case for profiling drush commands.

briantully:


shaal:
I tested these instructions, including running blackfire to profile drush commands.

It works GREAT!!! please merge so everyone can enjoy this easy setup.

PR #66, opened 2020-05-13
Add a git-exclude command

pcambra:
Often we find ourselves using ddev in environments where not necessarily everyone uses it and thus might not be appropriate to check the files into the repository.
Here's an initial version of a command that excludes all ddev files automatically.


rfay:
Rebased, minor edits, added to top-level README.

PR #67, opened 2020-05-15
solr: /var/solr/data has wrong permissions or wrong path

maddy2101:
When mounting the shipped core configurations and other resources provided by ext:solr, the mount point is /var/solr/data.

### DIFF
--- a/docker-compose-services/typo3-solr/docker-compose.solr.yaml
+++ b/docker-compose-services/typo3-solr/docker-compose.solr.yaml
@@ -14,10 +14,10 @@ services:
       - VIRTUAL_HOST=$DDEV_HOSTNAME
       - HTTP_EXPOSE=8983
     volumes:
-      - "./solr:/opt/solr/server/solr"
+      - "./solr:/opt/solr/data"
       # If you want your solr to persist over `ddev stop` and `ddev start` then uncomment the following line
       # If you uncomment it and want to flush your data you have to `ddev stop` and then
-      # `docker volume rm ddev-<projectname>_solrdata` to destroy it.
+      # `docker volume rm ddev-<projectname>_solr` to destroy it.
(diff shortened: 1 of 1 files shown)


rfay:
Please add testing instructions for this so that people can review it and verify it.

Also... your patch mounts onto /opt/solr/data, but your comment above says /var/solr/data

maddy2101:
will do, sorry for the hassle.

maddy2101:
How to test: Set up Solr as proposed in the readme. Call to the Solr UI (Your project-URL with :8983, see also ddev describe after starting).
Try and rename one of the displayed cores.
Before the change, you will get an error telling 
**Couldn't persist core properties to /var/solr/data/cores/english/core.properties : /var/solr/data/cores/english/core.properties**
In this example, I tried to rename the english core.

rfay:
However, your recipe as posted is using transient data in the container, not a mounted volume. 

AFAICT the problem with a mounted volume has nothing to do with docker versions, I tested with 2.1.0.5 and 2.2.0.5 as well as 2.3.0.2.
Basically, a new volume is created with root permissions, and the typo3solr/ext-solr image is built to run as the "solr" user, so it will never be able to access a new volume without some work to condition the volume.

To condition the volume, one would need to do something like this after creating it: `docker run -it  --rm -v ddev-<projectname>_solr:/tmp/solr busybox sh -c "chown -R 8983:8983 /tmp/solr"`

rfay:
The upstream solr docker image has the capability of setting environment variable SOLR_USER. So adding to the environment `- SOLR_USER=root` would make this work with docker volumes I believe.

rfay:
Closing this. I think the TYPO3 solr basically needs some work, and solr is tough.  No response from OP.

PR #68, opened 2020-05-19
feat(drupal7-subfolder): Add guide for a Drupal 7 subfolder site setup

klausi:
Moving this here from https://github.com/drud/ddev/issues/2268

Changed the guide a bit to already prepare the simple .ddev/nginx/subfolder.conf file. After https://github.com/drud/ddev/pull/2267 this should be one step simpler!


rfay:
Thanks!

recipes/drupal7-subfolder/README.md
```diff
@@ -0,0 +1,207 @@
+Writing a public summary of what we had to do to make DDEV work with a Drupal 7 site subfolder setup.
+
+## Subfolder
+
+A Drupal site might be operated from a subfolder (subdirectory) of a domain. Example for such a URL:
+* No Subfolder frontpage: https://example.com
+* Subfolder frontpage: https://example.com/jobs
+
+That means the Drupal site only lives in the `/jobs` part. A reverse proxy sits in front of the site and delivers requests either to the main page "/" or the Drupal site at "/jobs.
```
rfay, line 9: You need to explain the folder structure. I assume it's a main folder called "example.com" with a subdirectory called "jobs"

But where is the ddev config done? At the top level or below? (Where does the .ddev folder live?)

recipes/drupal7-subfolder/README.md
```diff
@@ -0,0 +1,207 @@
...
+Nginx needs 2 rewrite rules for the `/jobs` subfolder (replace /jobs with your subfolder name). Create `.ddev/nginx/subfolder.conf` (see also https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/#providing-custom-nginx-configuration )
+
+```
+# Rewrite the /jobs subfolder.
+# Special rule for the front page: we need to pass "/" explicitly here to
+# $_GET['q'] and PHP. This is not done by Nginx automatically and then
+# Drupal falls back to the full request URI which would include the
+# subfolder /jobs.
+rewrite ^/jobs/?$    /index.php?q=/&$args   last;
+rewrite ^/jobs/(.*)$   /$1   last;
```
rfay, line 26: With the current layout, it would be easier to just have this rewrite in the full nginx-site.conf.

recipes/drupal7-subfolder/README.md
```diff
@@ -0,0 +1,207 @@
...
+}
+```
+
+### Base URL
+
+We want the site to be reachable at https://example.ddev.site/jobs . We need to set the base URL for Drupal so achieve that. Add a settings.php copy hook in `.ddev/config.yaml`:
+
+```
+hooks:
+  pre-start:
```
rfay, line 174: I think a post-start hook that did a `drush vset` would be easier than this whole extra file?

recipes/drupal7-subfolder/README.md
```diff
@@ -0,0 +1,207 @@
...
+  - exec-host: cp -f sites/default/settings.ddev-custom.php sites/default/settings.php
```
rfay, line 175: If the site is in the top level directory (example.com) then this won't work... because the destination for the file is <subfolder>/sites/default/settings.php right?

rfay:
Since I didn't go forward with the PR to ddev, this one should be pretty easy to get going, just use a full nginx-site.conf, including the redirects

recipes/drupal7-subfolder/README.md
klausi: There is no file folder structure because we only want to run the /jobs subfolder site.

The ddev config is done at the usual root folder level.

recipes/drupal7-subfolder/README.md
klausi: I think that is not possible because we have custom code in the settings.php file.

recipes/drupal7-subfolder/README.md
klausi: Probably best if I make a complete drupal 7 repository with the .ddev config folder in it for reference. Then you can simply clone that and run ddev start.

rfay:
Thanks for the updates. I'm hoping to get back to backlog of ddev-contrib PRs in a week or two.

Note that the introduction of easier config of nginx-site.conf and apache-site.conf in v1.15.0 makes the config a lot easier, and there's a second-docroot example in both .ddev/nginx_full and .ddev/apache

klausi:
No worries, I will make some updates here in the meantime.

Demo Drupal 7 repository with correct DDEV config is ready: https://github.com/jobiqo/drupal7-subfolder

Will now work that into this guide.

klausi:
* Updated the guide with minor clarifications
* Mentioned the full Drupal 7 demo repository
* Updated for DDEV 1.15 nginx
* added dot.ddev folder an config files next to the guide README

Should be ready for review again!

recipes/drupal7-subfolder/README.md
klausi: yep, done!

recipes/drupal7-subfolder/README.md
klausi: it should work, the directory levels are the same as on a "normal" drupal site.

rfay:
Thanks so much for all your work on this @klausi ! I tested it and it works great.

The only remaining issue is that business of copying the ddev-subfolder.settings.php to sites/default/settings.php - 

* As far as I can tell, that's not necessary at all
* Copying into sites/default/settings.php doesn't do the right thing anyway, as it should be copying into jobs/sites/default

It didn't seem like it served any purpose to me. I missed it completely in your README and everything worked just fine.

Please see if you're OK with these changes. It's rebased, so you'll need to reset to your fork.

klausi:
Makes sense, thanks!

The settings file is done to catch requests to the main site. You solved that in your folder layout with an extra index.html file and an extra jobs folder. In our demo repository https://github.com/jobiqo/drupal7-subfolder we don't have that and rely on the settings.php snippet instead.

In our case the client's main site is completely out of our control (it is even on a different server).

But I don't care too much about that, I think the guide is good enough to be merged even in this version. Thank you!

rfay:
Thanks for all your work and patience on this!

klausi:
Same to you Randy :heart:

Issue #69, opened 2020-05-25
Mongodb suggestion: install php extension via webimage_extra_packages config option


binary-data:
The README suggests installing mongodb php extension by altering web image Dockerfile.

Although there is an alternative way: via config.yml. There is an option `webimage_extra_packages: [php-mongodb]`, which doesn't require to change Dockerfile. This might seem a bit cleaner.

Here is the reference: https://ddev.readthedocs.io/en/stable/users/extend/customizing-images/

rfay:
That's correct, do you want to do a PR? It does completely change the recipe, and you're right about the approach.

binary-data:
Sure. It shouldn't change the recipe dramatically though. And I'll need some time to actually test it.

PR #70, opened 2020-05-30
Change mongodb extension installation instructions

binary-data:
resolves drud/ddev-contrib#69

Installing MongoDB php extension via `webimage_extra_packages` config option instead of altering Dockerfile


rfay:
Shoot, you didn't mention the issue in the title, or mention this in the issue I don't think, so I redid it with more stuff. Could you please look at #83. #83 does add much more finesse to the docker-compose.mongo.yaml.

binary-data:
@rfay Looks good, thx!

rfay:
Thanks for taking a look!

PR #73, opened 2020-06-08
Setting up Behat with Selenium for Drupal 

erubino1977:


rfay:
Thanks, looking forward to testing this.  Do you mind if I edit a bit? It's not exactly in a matching format, but should be super useful. 

It looks like you have a couple of problems with signing the CLA. You can either add the email address you used to your github account, as in https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user or you can change the commit to use the one you want to use and then force-push again.  `git commit --amend --author="John Doe <john@doe.org>"`

rfay:
Thanks for this @erubino1977 - I got the time to test and edit, and added some commits. This has also been rebased to current, so you'll need to reset to your remote after `git fetch`. Anyway, please take a look and we'll get it in.

PR #74, opened 2020-06-08
Use stderr redirection to make "Import a sql dump if database is empty" quiet on start

mariohernandez:
* Updated script to redirect the stderr to /dev/null to avoid
showing an error on screen when target table is not found.

### DIFF
--- a/hook-examples/import-db-if-empty/import-if-empty.sh
+++ b/hook-examples/import-db-if-empty/import-if-empty.sh
@@ -1,9 +1,13 @@
 #!/bin/bash
 
-# Use a table that should exist in your database instead of "mytable"
-if ! mysql -e 'SELECT * FROM mytable;' db >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


hook-examples/import-db-if-empty/import-if-empty.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Use a table that should exist in your database instead of "mytable"
-if ! mysql -e 'SELECT * FROM mytable;' db > /dev/null; then
+if ! mysql -e 'SELECT * FROM mytable;' db 2>/dev/null; then
```
rfay, line 4:
```suggestion
if ! mysql -e 'SELECT * FROM mytable;' db >/dev/null 2>&1; then
```

rfay:
Rebased, improved the script a bit. Thanks for this!

PR #78, opened 2020-07-03
Remove env variable $WEBSERVER_DOCROOT usage

jonaseberle:
The env variable will not be there with 1.15
https://github.com/drud/ddev/issues/2330


rfay:
I think there are a couple of other usages in old_php - Could you push a commit to fix those as well?

rfay:
Thanks so much for this! Rebased, removed a couple more instances.

PR #81, opened 2020-07-28
Fix image names in "Using Old PHP Versions"

fjacobi:


rfay:
Did they change how the images are tagged? I tested this originally, and those images were correct then...

fjacobi:
I'm not really sure, but the images are named differently here. Perhaps they've streamlined them?

rfay:
Hmm, https://hub.docker.com/r/devilbox/php-fpm-5.2 is still there and appears to be maintained... ???

fjacobi:
Yes and if I test it today it works without any issues. 
Wonder what went wrong yesterday. /closed

PR #83, opened 2020-07-31
Simplify mongodb, use HTTPS_EXPOSE for access to mongo-express, fixes #69

### LINKED ISSUE #69: Mongodb suggestion: install php extension via webimage_extra_packages config option

rfay:
As proposed in #69, use simpler technique of getting php-mongo into the web container. Also exposes the mongo-express port in a way that works if more than one project uses it.

@wtfred could you please review this? Thanks!

### DIFF
--- a/docker-compose-services/mongodb/Dockerfile
+++ b/docker-compose-services/mongodb/Dockerfile
@@ -1,7 +1,0 @@
-# This is a Dockerfile for use with the ddev-webserver image
-# It should be in .ddev/web-build/Dockerfile
-# and then you can expand on it.
-ARG BASE_IMAGE=drud/ddev-webserver:v1.9.1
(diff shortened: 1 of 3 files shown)


wtfred:
@rfay Hey! A quick view of the changes looks fine, but i won't be able to try it before september!

rfay:
Hope you'll take a full look in September when you can, thanks!

PR #84, opened 2020-08-06
Experimental PHP8 recipe

rfay:


docker-compose-services/php8/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+DDEV-Local will directly support PHP8 as soon as possible, but not likely until the full release in late 2020. At that time this recipe will be obsolete.
+
+In the meantime, you can use the [official docker php image](https://hub.docker.com/_/php) (currently php:8.0.0alpha3-fpm-buster) to serve PHP.  This can be updated as new versions come out.
+
+1. Copy [docker-compose.php8.yaml](docker-compose.php8.yaml) to your project's .ddev folder.
+2. Copy [nginx-site.conf](nginx-site.conf) to your .ddev/nginx_full directory (overwriting the generated file there)
+3. `ddev start`
+
+You may find that your project has some problems with PHP8 or that PHP8 has some problems with your project :)
```
heddn, line 11: Not sure where to inject this, but the container name isn't web with this recipe. It is php. So to ssh into the container (for example) you have to `ddev ssh -s php`

rfay: There is a container named "web" and another one named "php", definitely need to mention that though.

rfay: I added a little mention. But both "web" and "php" are there, and the code is mounted into both in the same place. BUT, the php-fpm running on "web" is irrelevant and ignored.

heddn:
Also a note that composer and other things that you might assume would exist either. Its a plain docker container with really just php and FPM.

docker-compose-services/php8/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+DDEV-Local will directly support PHP8 as soon as possible, but not likely until the full release in late 2020. At that time this recipe will be obsolete.
+
+In the meantime, you can use the [official docker php image](https://hub.docker.com/_/php) (currently php:8.0.0alpha3-fpm-buster) to serve PHP.  This can be updated as new versions come out.
+
+1. Copy [docker-compose.php8.yaml](docker-compose.php8.yaml) to your project's .ddev folder.
+2. Copy [nginx-site.conf](nginx-site.conf) to your .ddev/nginx_full directory (overwriting the generated file there)
+3. `ddev start`
+
+Note that this introduces a container named `php`, which you can accesss via `ddev ssh -s php`. It has the exact same code mounted in the same place as in the `web` container, at `/var/www/html`.
```
heddn, line 11:
```suggestion
Note that this introduces a container named `php`, which you can accesss via `ddev ssh -s php`. It has the exact same code mounted in the same place as in the `web` container, at `/var/www/html`. It is just the vanilla upstream docker container, so things like composer, etc. all have to be manually installed.
```

rfay:
But everything is in the web container, and you can do all `ddev composer` or anything you'd ever expect to do as always. 

However, composer would be using php7.x (whatever you have configured).

docker-compose-services/php8/README.md
rfay: I wouldn't actually recommend to people to go into the php container though, so there's that. Why wouldn't you just use composer as always? The same code is mounted into both containers, can be changed on the host, in web container, or in php container. It's just being *interpreted* in the php container.

heddn:
I was testing composer v2 with php 8 and phpunit/phpunit. Right now, phpunit doesn't have 100% support for PHP 8. On <8, that same thing worked just fine. And depending on the version of PHP, composer will download different things.

rfay:
I'm certainly in favor of always using the same php with composer that is on the server side. However, what happens with composer 2 running with php 7.4 on the web container?

rfay:
You can use post-start hooks to install composer in web container, but you can't use Dockerfile since that's not supported except for web and db.

rfay:
Maybe we should give an example of how to install composer in web container?

heddn:
I just followed the instructions from https://getcomposer.org/download/ to install composer. To get v2, I did `php composer.phar selfupdate --preview`

rfay:
Added hook example for composer install, including composer 2. Couldn't get a composer install to work on D9.0.2 though.

PR #89, opened 2020-09-04
Set default language (or other settings) in phpmyadmin

Tuurlijk:


Tuurlijk:
Submission Contains Third Party Materials

rfay:
I'm not sure what you're saying or referring to.

Tuurlijk:
I'm referring to the paragraph:

6. Third Party Contributions.

Should You wish to submit work that is not Your original creation, You may submit it to Drud separately from any Contribution, identifying the complete details of its source and of any license or other restriction (including, but not limited to, related patents, trademarks, and license agreements) of which you are aware, retain all copyright and license information that was in the materials as You received them, and conspicuously marking the work as "Submission Contains Third Party Materials."

rfay:
That would be the situation where you were submitting somebody else's work, for example if you were adding a PR that added a third-party project directly into the codebase. You're obviously not doing that.

Tuurlijk:
So what happens if I don't sign? I don't understand paragraph 4. `Patent License`. The language is too complicated.

I wish for my contribution to remain 'visible' and not be sold with the complete company / product. It needs to remain open-source.

rfay:
Maybe we should get rid of that or the whole thing. I think it's normal stuff, but since nothing here can be sold, seems odd.

rfay:
So if you were to take your "contribution" to closed-source, it would then not be "visible" in ddev-contrib as a contribution by you.  Not sure how any of those things could happen. But it's trying to prevent the same thing, people doing a PR of something that actually lives elsewhere, and that might not continue to be open source.

rfay:
If you're put off by the CLA, it would be fine to put this on Stack Overflow. Overall, we just really appreciate your contribution and following up on this. Apparently lawyers have their reasons.

rfay (approved):

PR #91, opened 2020-09-27
Add a new service with Portainer

davidjguru:
Sometimes, many people are not comfortable enough using the command line or just feel better using graphical user interfaces (GUIs). Due to this, the example committed here adds a new case about extending DDEV-Local networking by adding a new container for a Portainer service, the visual tool for management of containers.


rfay:
Nice! 

Are you using macOS? Have you looked at the new container visualization built in to the latest (edge) docker desktop?

rfay:
One thing I haven't every quite understood is why people need to see things at the container level like they do with portainer. Every time I fiddle with it I think "Why do I want to see the containers?". Why do you want to see the containers?

davidjguru:
Hi @rfay,
 I'm not a macOS user, usually I'm a Linux (Ubuntu / Debian) user. I don't know how Docker Desktop, works, sorry. 
Well, the truth is I don't really need to "see" the containers. But on a daily basis, training people with a "_junior_" profile, I have observed that is more comfortable thinking in containers like items that they can "see". 

I made this case as a training tool, as a didactic resource. There're so many people with some (initial) panic over the prompt. Facilitating and gradually advancing people is part of my job. This is my context. 

Greetings, 
David.

rfay:
In general, my goal with ddev is for people to "see" ddev, and not docker or its containers or the complexity involved there. That's why I ask. 

Of course, the need for a gui for ddev is a longstanding concern. We once had one, but it wasn't really maintainable.

rfay:
Thanks for this!

PR #92, opened 2020-10-01
Drop the database before importing

pcambra:
I think we do this in the mysql counterpart, this might be useful for someone.


rfay (approved):
Looks good to me. mariadb and mysql (built-in) drop by default but have the option of not doing so, but dropping first is the most reliable for most uses.

PR #97, opened 2020-10-21
Create `drush uli` command

mglaman:


custom-commands/drush-uli/uli
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+## Description: Log into the Drupal site with a one-time login link.
+URL=$(ddev exec drush uli | tr -d '\r')
+
+case $OSTYPE in
```
mglaman, line 5: The `launch` host command always prepends the DDEV_URL, so we cannot shortcut to the following

```
ddev launch $(ddev exec drush uli)
```

rfay:
`ddev drush` is already in current alphas. You'll like it.

mglaman:
I know. But that's an extra word. And does `ddev drush uli` launch the browser, though? I thought it didn't due to executing Drush from inside the container

rfay:
`ddev drush uli` outputs the URL. It's the user's job to click it.

rfay:
Nice to have you here!

mglaman:
We can decline this then, I'll keep it for myself :). I've started writing more, but I'll see what is more generic and useful.

rfay:
It would be super cool to do what drush does and pop the browser using terminal characters instead of the "open" or whatever technique `ddev launch` uses.

mglaman:
I've got an idea for the regular Drush command support to do that. I'll open a feature request later

PR #98, opened 2020-10-22
Add BROWSERTEST_OUTPUT_BASE_URL to env vars in drupalci-crhomedriver

eojthebrave:
## Problem

URLs output by phpunit pointing to test artifacts use http://web for the address but should use the URL of the web container so they can be viewed.

## Resolution

This adds the `BROWSERTEST_OUTPUT_BASE_URL` to the environment variables in the web container. Which, phpunit will use when outputting links to saved HTML assets generated during a test run.

Before:

html 2020-10-22 15-19-22

After:

html 2020-10-22 16-41-26

### DIFF
--- a/docker-compose-services/drupalci-chromedriver/docker-compose.chromedriver.yaml
+++ b/docker-compose-services/drupalci-chromedriver/docker-compose.chromedriver.yaml
@@ -17,5 +17,5 @@ services:
       # - SIMPLETEST_DB=mysql://db:db@db:3306/db
       - SIMPLETEST_BASE_URL=http://web
       - BROWSERTEST_OUTPUT_DIRECTORY=/tmp
-      - BROWSERTEST_OUTPUT_BASE_URL=http://${DDEV_SITENAME}.ddev.site
+      - BROWSERTEST_OUTPUT_BASE_URL=${DDEV_PRIMARY_URL}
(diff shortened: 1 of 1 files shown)


docker-compose-services/drupalci-chromedriver/docker-compose.chromedriver.yaml
```diff
@@ -17,4 +17,5 @@ services:
       # - SIMPLETEST_DB=mysql://db:db@db:3306/db
       - SIMPLETEST_BASE_URL=http://web
       - BROWSERTEST_OUTPUT_DIRECTORY=/tmp
+      - BROWSERTEST_OUTPUT_BASE_URL=http://${DDEV_SITENAME}.ddev.site
```
rfay, line 20: I think this would be more likely to work with $DDEV_PRIMARY_URL, please try it out and if it's good, you can commit here with the button and I'll pull it.
```suggestion
      - BROWSERTEST_OUTPUT_BASE_URL=${DDEV_PRIMARY_URL}
```

rfay (changes requested):
Please try ${DDEV_PRIMARY_URL}, thanks!

docker-compose-services/drupalci-chromedriver/docker-compose.chromedriver.yaml
eojthebrave: Just tested and that works great. Thanks for the suggestion @rfay.

rfay (approved):

PR #99, opened 2020-10-23
Change Portainer port on the host

davidjguru:
Change port for the Host in order to avoid conflicts when xdebug is enabled. 
Now uses port 9001, mapping 9001:9000 for Portainer.

### DIFF
--- a/docker-compose-services/portainer/README.md
+++ b/docker-compose-services/portainer/README.md
@@ -23,26 +23,39 @@ Portainer is a visual mode for management of containers and networks. Is open-so
 
 ## Connection
 
-Ok, by default you can access to the Portainer interface at the next address from your browser:
+Ok, by default you can access to the Portainer interface at the next addresses from your browser:
 
 ```
-https://projectname.ddev.site:9001
-https://127.0.0.1:9001
+# using http
+http://projectname.ddev.site:9001
+http://127.0.0.1:9001
+
+# or using https
+https://projectname.ddev.site:8001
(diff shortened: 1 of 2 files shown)


rfay (changes requested):
If portainer is running http (I think it is) then you'd be a lot better using HTTP_EXPOSE (and `expose` instead of `ports`. 

Currently this could only run on a single project. 

Several other examples here use the environment variable HTTP_EXPOSE and HTTPS_EXPOSE to expose something that is then routed through the ddev-router, meaning you can run multiple projects with this at once.

davidjguru:
Hi @rfay,
I have made modifications on the file taking into account your notes. I have tried to deploy several projects in parallel and everything seems to go very well with this port mapping for Portainer, working as expected. 
Thanks a lot for your feedback. 
Greetings.

docker-compose-services/portainer/README.md
```diff
@@ -26,21 +26,21 @@ Portainer is a visual mode for management of containers and networks. Is open-so
...
 
 Just do:
 
 ```
-ddev stop
-vim docker-compose.portainer.yaml
+$ ddev stop
+$ vim docker-compose.portainer.yaml
 ports:
   - '3434:9000'
```
rfay, line 41: It looks like you still need to change this bit.

rfay (changes requested):

rfay (approved):
Thanks so much for all your good work on this!

PR #100, opened 2020-11-13
WIP: Cups docker-compose recipe

AronNovak:


rfay:
print server!

rfay:
Converted this to draft. Are you still interested in pursuing it? Can't think of a reason in the world to have a print server in ddev config!

AronNovak:
@rfay We had a use-case, but printing inside Docker is more painful as we perceived originally, it's heavily platform-dependent, just like as it would be native.

PR #106, opened 2020-12-23
[WIP] Add custom command "general_log"

jonaseberle:


rfay (changes requested):
It seems like query-log would be a better name for the command than general_log?

jonaseberle:
@rfay fine with me of course. I just chose it because the MySQL setting and feature is general_log.

rfay:
I was just thinking about a command name that would imply it had to do with database stuff. In the context of database, general_log is explanatory, but in the context of a full web environment like ddev, it could mean so many different things.

jonaseberle:
Name suggestion: `db-log`

[WIP] Looking into integrating slow query log

cmuench:
> Name suggestion: `db-log`
> 
> [WIP] Looking into integrating slow query log

Why not a general "log" command with sub-commands?

Example:

`ddev log db`

Then we are able to extend the command by new loggers without a too long command list.

jonaseberle:
@cmuench I wouldn't want to override a stock command (which is not possible btw). I think of ddev-contrib as something like Steam Greenlight where ideas can be shared and vetted. If really helpful, they might maybe eventually trigger something upstream.

rfay:
This is a great contribution, hoping you can get back to it @jonaseberle, thanks!

rfay (approved):
Thanks!

PR #109, opened 2020-12-30
Change elasticsearch version from 7.6.1 to 7.10.1

nickw108:
The official elasticsearch dockerhub only lists the tags 7.10.1 and 6.8.13 as supported. When I ran ddev start with tag 7.6.1 I got an error "pulling elasticsearch (elasticsearch:7.6.1)... no matching manifest for linux/arm64/v8 in the manifest list entries". With tag 7.10.1 it worked for me.

### DIFF
--- a/docker-compose-services/elasticsearch/docker-compose.elasticsearch.yaml
+++ b/docker-compose-services/elasticsearch/docker-compose.elasticsearch.yaml
@@ -3,7 +3,7 @@ services:
   elasticsearch:
     container_name: ddev-${DDEV_SITENAME}-elasticsearch
     hostname: ${DDEV_SITENAME}-elasticsearch
-    image: elasticsearch:7.6.1
+    image: elasticsearch:7.10.1
     ports:
       - "9200"
       - "9300"


nickw108:
Just checked the Elasticsearch [supported versions documentation](https://devdocs.magento.com/guides/v2.4/install-gde/prereq/elasticsearch.html#es-spt-versions) for Magento 2.4 🤔 
> Magento 2.4.x is tested with Elasticsearch 7.6.x only. You can use other 7.x versions at your discretion, but we recommend using the tested version of Elasticsearch.

rfay:
It looks like they only recently started pushing arm64 images, see https://hub.docker.com/_/elasticsearch?tab=tags&page=1&ordering=last_updated

So you'll likely need to use the newer version on your arm64 machine, but it doesn't seem like we'd want to change it based on your research? 

BTW, on what type of machine are you using arm64? Mac M1? How is it going?

rfay (approved):

nickw108:
> So you'll likely need to use the newer version on your arm64 machine, but it doesn't seem like we'd want to change it based on your research?

Makes sense, confused that you merged my pull request 🤓 

> BTW, on what type of machine are you using arm64? Mac M1? How is it going?

Yeah, on Mac M1. It's running, not as fast as I hoped, though. I saw [macOS NFS Setup](https://ddev.readthedocs.io/en/stable/users/performance/#macos-nfs-setup) in the documentation, which I might try. But I was wondering if there's something like Docker volume caching available, too?

rfay:
I pulled it because you were interested, didn't seem to do any harm, and because @cmuench tweeted that Elasticsearch 7.10 will be supported in Magento 2.4.2, https://twitter.com/cmuench/status/1344689116051550209

You absolutely should do NFS, it makes a huge difference. There isn't currently a good caching option for docker. But Magento2 is absolutely a hog, plays to docker's worst characteristics.

PR #110, opened 2021-01-09
inotify-proxy to enable file watchers on NFS shares

cmuench:
This command is used at netz98 to enable all frontend developers file watch tasks.
It runs fine with grunt, Magento 2 Cache Clean, less, sass, typescript and it's faster than other similar solutions.
If you work with ddev and Magento 2 on a Mac, it's a must have.


custom-commands/inotify-proxy/README.md
```diff
@@ -0,0 +1,11 @@
+# inotify-proxy to enable file watchers on NFS shares
+
+This tools helps to detect changed files in Docker Containers if NFS mound is used.
```
rfay, line 3: Please say where the custom command is to be installed.

rfay: It's not clear reading the inotify-proxy project page whether this is a tool that runs on the host side or the container side. Or if it's linux only for the host side. If it's linux-only, please clear that up there and here. Also, if it's linux-only, could you add an arm64 build?

rfay (changes requested):
What a great initiative! If this is general-purpose lots of people will want it. @cweagans you'll want to take a look at this too.

custom-commands/inotify-proxy/README.md
cmuench: Right... I am a bit professionally blinkered. I will update the docs.
It's running inside the container. I can add an ARM build but I have no device to test it.

cmuench: Added additional description and released v2.1.0 of inotify-proxy with linux arm64 support.

cweagans:
This looks really cool! Thanks for putting this together!

If I'm reading https://github.com/cmuench/inotify-proxy/blob/master/internal/watcher/watcher.go#L93-L107 correctly, this wouldn't be able to emit inotify events for deletions, right? Still super useful though.

Also, we've had problems in the past with any kind of polling inside of the container. It would either just die at some point, not see changes/created/delete files until a long delay elapsed, or other strange behavior. If I'm understanding inotify-proxy correctly, it's polling all of the configured paths for changes, so that might be something to watch out for.

One alternative approach here (if you were feeling ambitious 🙂 ) would be to use something like https://github.com/fsnotify/fsnotify on the host and relay those notifications into the guest over a network connection or something. That way, you're more likely to get _all_ of the FS changes. Of course, that means that setup is a little more complex for end-users.

cmuench:
@cweagans Thanks for your review. I am stiff learning golang. So this project and ddev help me to create more than "hello-world" applications.
I never missed that delete events are not trigged but I will check if we can get them, too. We also tried, in our agency, several other tools. I did not plan to reinvent the wheel. At the end the other tools had to much downsides. The most popular node package is compiling C code in the background. That requires build-tools installed in your ddev project and it has a memory leak. There is no garbage collection. That's why the projects are getting slower and slower. 
My tool fixed that. In the meantime this tool is used in all our ddev projects.

If I understand fsnotify apporach ... it runs on the host and sends notifications via network into the container.
That's maybe a solution which works in every case, but it sound also more complicated.

cmuench:
@cweagans Checked the fsnotify library. Seems that there is no go library which sends the IN_DELETE syscall. They all use the `golang.org/x/sys/unix` package which provides only inotify watch logic. 

I will try to send a IN_DELETE (or IN_DELETE_SELF ?) syscall in my tool.
reference: https://github.com/torvalds/linux/blob/master/include/uapi/linux/inotify.h#L38

cmuench:
@cweagans It seems that the kernel is blocking my syscall... 
Got: `2021/01/11 11:13:00 error: operation not permitted`
Interesting.

cweagans:
@cmuench you cannot manually emit inotify events like that. There's no userspace interface to do that. missing delete events is probably not that big of a deal :)

cweagans:
(also, just to clarify: none of what i posted should be taken as a blocker for getting this in. I don't work for ddev anymore and I'm not speaking as a maintainer, just as an interested 3rd party)

custom-commands/inotify-proxy/README.md
```diff
@@ -0,0 +1,23 @@
+# inotify-proxy to enable file watchers on NFS shares
+
```
rfay, line 2: A little introduction will probably help here, since people don't automatically know what this is about.
```suggestion

On a regular Linux or macOS filesystem, a process can subscribe to notifications about file changes ("inotify" or "fsnotify").  Most IDEs like PhpStorm depend on this notification to update files or rendering when they change. However, the NFS protocol predates the inotify feature, and so does not support it. However, most developers on macOS and Windows use the [nfs_mount_enabled feature](https://ddev.readthedocs.io/en/stable/users/performance/#using-nfs-to-mount-the-project-into-the-web-container) to mount files from the host into the container instead of using the docker-native technique. Performance is much improved, but of course NFS does not support notifications.

(Note that this is not necessary on Linux hosts; NFS is not needed there because the Docker filesystem mount is so fast.)
```

rfay: BTW, does this work on WIndows? What's the easiest way to test?

custom-commands/inotify-proxy/README.md
```diff
@@ -0,0 +1,23 @@
...
+## Installation
+
+Copy the file `inotify` into your project **.ddev/commands/web** directory.
+
+## Run the command
+
+```sh
+ddev inotify
+```
+
```
rfay, line 22: Please add a `## How to test` to demonstrate that it's working in a simple way.

rfay (changes requested):
It's awesome, of course. Thanks for the great contribution. I have a few trivial requests and some requests for additional understanding.

custom-commands/inotify-proxy/README.md
cmuench: It's designed to work inside a docker container. IMHO it's only running on linux.

rfay: I understand, but does this work on a windows host?

cmuench: Never tested that, because we do not work with Windows. 
Theoretical It should also work on a Windows host with NFS share, but I tested it only with MacOS and Linux.

rfay: I'll test it if I know how to test it. I think one of those other tools has a host-side listener that worked. Still not understanding how this works without a host-side listener.

cmuench: It has a internal dict with the stats of each watched file.
In every watch cycle it checks if the file is changed. If it's changed, the file will be touched which triggers a Inotify event.
It's only necessary to start the tool by this command after "ddev start".

rfay: So it works by changing the timestamp of the file? That's a pretty slick, for generated files. Seems like a simpler technique than some others.

cmuench: That's the whole magic.
https://github.com/cmuench/inotify-proxy/blob/1fba6da45ab75a77867b22b3f88d5170ad04ff78/internal/watcher/watcher.go#L98

rfay: I guess that's what I don't understand about the whole process. If a file changes on the container filesystem, its timestamp already changes. And this technique *still* uses NFS to provide that change to the host filesystem.

cmuench: right :-)

cweagans: @rfay writes to NFS originating on the machine that has mounted an NFS export (in this case, the container) still trigger inotify events as expected. Events originating on the host that's providing the NFS export (the host mac system) don't.

@cmuench have you been using this for long periods of time with large projects? This is effectively just polling NFS for updated metadata and we've run into problems with that approach when using e.g. the polling mode in webpack. I'm curious about the overall reliability here. If you change 1000 files on the host, do you get 1000 inotify events in the guest? Same for 10k?

rfay: I guess people use inotify (and miss it) two different ways. One is changes in the container not resulting in inotify on the host (which I don't think this does anything about) and the other is changes on the host not resulting in inotify to the container (which this does, right?)

cmuench: @cweagans At netz98 the frontend developers use this tools for different file watch tasks in Magento 2 shops.
Is there any php system with more files then Magento 2? 😉
We do not have the use case that there are several thousand files changing at once. The tool has a garbage collector to prevent a memory leak.
My tool is the golang version if this NPM package https://www.npmjs.com/package/inotify-proxy
The NPM package compiles a c program.
If we change 1.000 files, there should be 1.000 change events in the guest. Currently we do not have problems. The tool is also used by other agencies in the Magento eco-system.

cmuench: @rfay If you work directly inside a container then you do not have problems with missing inotify events. The issue is that someone works in e.g. PhpStorm and changes files. The watch task runs inside the container. If NFS used, no inotify event is triggered. Without the inotify event, the watch task inside the container is not seeing any change. That's annoying.
In the Magento world the people use less/sass/grunt and the Magento Cache Cleaner tool.
Without a good solution, ddev has as low acceptance and people work directly on the host machine.

rfay: Right, but the opposite problem people have is with GUIs that are expecting notifications when running on the host, and doing a build that runs in the container. So for example, they have gulp or compass running in the container and they change css files.

rfay: I guess what we need to do here is make sure we're clear which problem is being solved. And pioneer best practices for solving it. Because the more you can do your work where things work best (inside the container, Linux) the better off you are.

rfay (approved):
Thanks!

PR #118, opened 2021-02-04
Small clarifications for PHPUnit/Chromedriver setup and possible issues.

damienmckenna:
This helps clarify a few items I ran into while trying to get PHPUnit to work. Thank you for the documentation!

### DIFF
--- a/docker-compose-services/drupalci-chromedriver/README.md
+++ b/docker-compose-services/drupalci-chromedriver/README.md
@@ -14,7 +14,7 @@ It is based on [Matt Glaman's excelent blog post](https://glamanate.com/blog/run
 * Ensure that one of the two `SIMPLETEST_DB` options is uncommented.
 * Start (or restart) DDEV to have the service initialized: `ddev start`
 * Your Drupal project must be built with --require-dev to get necessary dependencies like phpunit (for example, `ddev composer create drupal/recommended-project --require-dev`)
-* To test the setup, `ddev ssh` and run a Drupal unit test that triggers ChromeDriver. Example:
+* To test the setup, `ddev ssh` and run a Drupal unit test that triggers ChromeDriver. Example (assumes Drupal core is in the `web` directory):
 
 ```bash
 user@demo-web:/var/www/html$ phpunit --verbose -c web/core/phpunit.xml.dist web/core/modules/system/tests/src/FunctionalJavascript/System/DateFormatTest.php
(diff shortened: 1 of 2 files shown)


docker-compose-services/drupalci-chromedriver/README.md
```diff
@@ -19,5 +20,10 @@ It is based on [Matt Glaman's excelent blog post](https://glamanate.com/blog/run
 user@demo-web:/var/www/html$ phpunit --verbose -c web/core/phpunit.xml.dist web/core/modules/system/tests/src/FunctionalJavascript/System/DateFormatTest.php
 ```
 
+## Notes
+
+* In order for this to work all other ddev instances must be turned off, only the Drupal core instance which will be used for the PHPUnit tests can be running.
```
rfay, line 26: There's no reason this should be the case. What were you using for SIMPLETEST_BASE_URL?

rfay: Oh, I see the problem. The chromedriver service needs 
```
links:
  web:web
```

The problem is described in https://github.com/drud/ddev/issues/2744 but it's an easy fix

docker-compose-services/drupalci-chromedriver/docker-compose.chromedriver.yaml
```diff
@@ -1,4 +1,8 @@
 ---
+# Docker-ChromeDriver configuration for DDEV-Local.
+# There is one item that must be configured in this file in order for it to
+# work, please see below.
+
 version: '3.6'
 services:
   chromedriver:
```
rfay, line 8:
```suggestion
  chromedriver:
      links:
          web:web
```

docker-compose-services/drupalci-chromedriver/docker-compose.chromedriver.yaml
```diff
@@ -12,9 +16,13 @@ services:
...
     environment:
-      # Pick if you want to run tests as sqlite or mysql and uncomment.
+      # *** One of these must be uncommented ***
+      # In order for the system to work, one of these must be uncommented so
+      # that the test system can connect to the database.
       # - SIMPLETEST_DB=sqlite://tmp/test.sqlite
       # - SIMPLETEST_DB=mysql://db:db@db:3306/db
+      #
+      # Note: Do not modify the base URL value.
       - SIMPLETEST_BASE_URL=http://web
```
rfay, line 26: The suggestion above makes "web" be unambiguous, so this recipe can work with other projects running.

rfay (changes requested):
Thanks for battling this through. I'd appreciate if you could test the suggestion I made. I think it will solve the problem of having to stop other projects.

docker-compose-services/drupalci-chromedriver/README.md
damienmckenna: When I add "web:web" to the links list it gives this error:

`A service can not link to itself: web`

damienmckenna: Never mind, I added the link to the wrong service.

damienmckenna: I added the web:web link to the chromedriver services section so it now looks like this:

```
services:
  chromedriver:
    image:  drupalci/webdriver-chromedriver:production
    container_name: ddev-${DDEV_SITENAME}-chromedriver
    labels:
      com.ddev.site-name: ${DDEV_SITENAME}
      com.ddev.approot: $DDEV_APPROOT
    links:
      - web:web
```
However it now gives this error:
`Circular dependency between chromedriver and web and db `

rfay: Thanks for exploring this so carefully. The issue is that "web" may be ambiguous, and "chromedriver" will be ambiguous if there are multiple projects running at the same time.  However, both mostly work without the links. But the addition of web in chromedriver will help. The one in web is less important. 

Here's what I recommend, https://gist.github.com/rfay/801d382cf08bebf4811b43ff0ea8fa31

This 
* changes the name of the service to "chrome", so we can use the hostname "chrome" instead of switching between "chromedriver" and "chrome"
* Uses "external links" for the chrome container to address <hostname>.ddev.site. This routes traffic through the router, which is probably a better test anyway. It *could* also point to "web" instead of ddev-router. 

I think this will probably do it. If it doesn't work right away for you, maybe we could work together a little bit in slack and get it nailed?

rfay (approved):

PR #121, opened 2021-03-04
Replace `sync` with `platform-sync` in Usage

boulaffasae:


rfay:
Actually... I'd really like it if you'd test out the new platform support in ddev-local alphas. Use the latest alpha, https://github.com/drud/ddev/releases/tag/v1.17.0-alpha6 but the info about support and links to docs are in https://github.com/drud/ddev/releases/tag/v1.17.0-alpha5

I think they'll make this obsolete.

PR #124, opened 2021-03-18
Update command to allow Redis to correctly open without error and use latest version of Redis.

rahaddon:
Change command to allow Redis to correctly open without error and use latest Redis version. Note: Required to add `REDIS_HOST=redis` to Laravel projects .env file as default of localhost is not appropriate inside the container.

### DIFF
--- a/docker-compose-services/redis/docker-compose.redis.yaml
+++ b/docker-compose-services/redis/docker-compose.redis.yaml
@@ -5,7 +5,7 @@ version: "3.6"
 services:
   redis:
     container_name: ddev-${DDEV_SITENAME}-redis
-    image: redis:4
+    image: redis:latest
     ports:
       - 6379
     labels:
@@ -14,7 +14,7 @@ services:
     volumes:
       - ".:/mnt/ddev_config"
       - "./redis:/usr/local/etc/redis"
-    command: ["redis-server", "/usr/local/etc/redis/redis.conf"]
+    command: ["redis-server", "--bind", "redis", "--port", "6379"]
   web:
     depends_on:
       - redis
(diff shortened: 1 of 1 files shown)


docker-compose-services/redis/docker-compose.redis.yaml
```diff
@@ -5,7 +5,7 @@ version: "3.6"
 services:
   redis:
     container_name: ddev-${DDEV_SITENAME}-redis
-    image: redis:4
+    image: redis:latest
```
rfay, line 8: Using "latest" is always a bad idea with docker images.  I think looking at https://hub.docker.com/_/redis?tab=description&page=1&ordering=last_updated what you want is `redis: 6.2`
```suggestion
    image: redis:6.2
```

docker-compose-services/redis/docker-compose.redis.yaml
```diff
@@ -14,7 +14,7 @@ services:
     volumes:
       - ".:/mnt/ddev_config"
       - "./redis:/usr/local/etc/redis"
-    command: ["redis-server", "/usr/local/etc/redis/redis.conf"]
+    command: ["redis-server", "--bind", "redis", "--port", "6379"]
```
rfay, line 17: This change throws away the redis config that is provided by the user, which is mounted in /usr/local/etc/resdis.

rfay (changes requested):
Simple change to redis image, but the problem with not using the provided redis.conf is a problem that you need to figure out.

docker-compose-services/redis/docker-compose.redis.yaml
rahaddon: I can confirm it as working using redis:6.2

rfay:
Although you're now using a specific docker tag, the change here means that provided configuration is not used.

rfay:
No response on this after a couple of months, so I'm going to close it. Thanks for taking the time to make the initial changes, but I don't think they're correct. If you want to discuss more, feel free to comment here, and I'll be happy to reopen this.

PR #93, opened 2020-10-02
Add recipe for Solr 4

stborchert:
For a (Drupal) project we needed to integrate _Apache Solr_ version 4 in our DDEV setup.
The recipe shows a simple example on how to run _Solr_ version 4 in DDEV.

PR #125, opened 2021-04-15
Add recipe for Solr 7

becw:
I recently tried to use Solr 7 following the [setup instructions in the ddev documentation](https://ddev.readthedocs.io/en/stable/users/extend/additional-services/#apache-solr). I found that the default docker-compose and core update script didn't work with the Solr 7 docker image, because some of the paths were different.

It's important to be able to match the Solr version on local development environments to what's available on hosting platforms, and at this time Acquia offers Solr 7.

### Testing instructions

1. Copy the `docker-compose.solr.yaml` file and the `solr/` directory into your `.ddev/` directory
2. Copy your desired solr core configuration into the `.ddev/solr/conf/` directory
  * You may be able to download your hosting provider's default core configuration
  * You can also use the [jump start configuration](https://git.drupalcode.org/project/search_api_solr/-/tree/4.x/jump-start/solr7/config-set) from `search_api_solr` module
3. Start (or restart) ddev
4. Visit the Solr web UI for your ddev setup, like `http://EXAMPLE.ddev.site:8983`:
  * You should not see any errors at the top of the page
  * In the center of the page, it should show the Solr version as 7 (currently 7.7.3)
  * In the left sidebar, the "Core Selector" dropdown should contain the core `dev`

image

### DIFF
--- a/docker-compose-services/solr-7/README.md
+++ b/docker-compose-services/solr-7/README.md
@@ -33,3 +33,5 @@ services.solr:
 ```
 
 You'll also need to use the updated `solr-configupdate.sh` script, because the Solr 7 image uses the directory `/opt/solr/server/solr/mycores/${CORENAME}/conf`, which is different from some other versions.
+
+**Contributed by [@becw](https://github.com/becw)**
--- a/docker-compose-services/solr-7/README.md
+++ b/docker-compose-services/solr-7/README.md
@@ -1,0 +1,35 @@
+# Solr 7 for DDEV
+
+The documented ApacheSolr integration with DDEV assumes version 8 of Solr, but the Solr docker images have different configuration locations between versions of Solr. This means that the docker-compose file and `solr-configupdate.sh` script need to be tweaked for use with Solr 7.
+
+## General documentation
+
+* [DDEV Integration documentation](https://ddev.readthedocs.io/en/stable/users/extend/additional-services/#apache-solr)
+* [Docker Solr documentation](https://hub.docker.com/_/solr/)
+
+## Installation
+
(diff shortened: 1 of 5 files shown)


rfay:
Is this a duplicate of https://github.com/drud/ddev-contrib/pull/93 ?

becw:
@rfay Nope. This one is for Solr 7, not Solr 4.

rfay:
Please take a look at the other one. I think both of yours probably cover Solr 4-7

becw:
@rfay Nope, they're actually quite different. This is also different from the solr-5 recipe, which is where I started from.

rfay:
Please add to the top-level readme as well

rfay (changes requested):
Looking forward to getting this in. Please add manual testing instructions and add it to the top-level README.

docker-compose-services/solr-7/README.md
```diff
@@ -0,0 +1,35 @@
...
+If you don't want to follow the installation steps above, you can make these changes manually:
+
+```
+services.solr:
+  image: solr:7
+  volumes:
+    - solr:/opt/solr
+```
+
+You'll also need to use the updated `solr-configupdate.sh` script, because the Solr 7 image uses the directory `/opt/solr/server/solr/mycores/${CORENAME}/conf`, which is different from some other versions.
```
rfay, line 35:
```suggestion
You'll also need to use the updated `solr-configupdate.sh` script, because the Solr 7 image uses the directory `/opt/solr/server/solr/mycores/${CORENAME}/conf`, which is different from some other versions.

**Contributed by [@becw](https://github.com/becw)**
```

rfay:
Thanks for this!

PR #126, opened 2021-04-17
Added pgsql command for restoring a pg custom format dump.

penyaskito:


rfay (changes requested):
Thanks for this! Please add a a bit to the postgres README explaining that this is there and how to use it. Thanks!

penyaskito:
Thanks Randy! Added some bits to the README. Also found out that I didn't put the command in the right folder, so moved it.

rfay (approved):
Thanks for this!

PR #135, opened 2021-05-31
Fix location for general_log file

### LINKED ISSUE #133: general_log script is not working
I have copied the file general_log into my project's .ddev/commands/db and then I have restarted ddev. Afterwards I run following command in my project folder 

`ddev general_log`

I got the following message

```

Enabling general_log mysql: [Warning] Using a password on the command line interface can be insecure.
✓
Press CTRL+C to exit.

Failed to run general_log : exit status 1


```

jonaseberle:

The default `general_log_file` settings takes the DB container hostname into account.

This change just sets that setting to a known location that we can use later.

Run `ddev general_log` on different projects (with and without project names beginning with `www.`). You should not see an error.

## Related Issue Link(s):

Fixes #133

### DIFF
--- a/custom-commands/general-log/general_log
+++ b/custom-commands/general-log/general_log
@@ -59,4 +59,4 @@ function shutdown() {
 trap shutdown INT
 
 enable
-tail -f /var/lib/mysql/general_log.log 2> /dev/null
+tail -f /var/lib/mysql/general_log.log 
--- a/custom-commands/general-log/general_log
+++ b/custom-commands/general-log/general_log
@@ -42,6 +42,7 @@ function spinner() {
 }
 function enable() {
     printf "Enabling general_log "
+    mysql -uroot -proot <<< "SET GLOBAL general_log_file='general_log.log'"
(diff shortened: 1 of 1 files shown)


custom-commands/general-log/general_log
```diff
@@ -58,4 +59,4 @@ function shutdown() {
 trap shutdown INT
 
 enable
-tail -f /var/lib/mysql/www.log 2> /dev/null
+tail -f /var/lib/mysql/general_log.log 2> /dev/null
```
rfay, line 62: I don't see any reason to redirect stderr to /dev/null - just confuses people if an error should occur. Not redirecting would have made for an understandable error in the bug this is fixing.

rfay (changes requested):
Thanks for the quick fix! Just want to remove the stderr redirection if there's no good reason for it.

custom-commands/general-log/general_log
jonaseberle: yes, good idea.

rfay (approved):
Thanks!

PR #137, opened 2021-06-14
Renamed the solr4 directory

damienmckenna:
Other Solr directories were named with a dash, e.g. "solr-5", so this renames "solr4" to "solr-4" and makes them all display together in the correct order.


rfay (changes requested):
You'll have to change the top-level readme as well, right?

damienmckenna:
> You'll have to change the top-level readme as well, right?

Ah, forgot about that one. Updated.

rfay (approved):
Thanks. I imagine you'll now have to rebase the other PR.

PR #140, opened 2021-06-14
Troubleshooting tips for Solr.

damienmckenna:
This provides some tips for troubleshooting Solr.

### DIFF
--- a/docker-compose-services/solr-5/README.md
+++ b/docker-compose-services/solr-5/README.md
@@ -83,7 +83,7 @@ e.g. `http://myproject.ddev.site:8983`.
 
 * The brower may redirect to https://myproject.ddev.site:8983. This is a
   default behavior in Safari. The solution is to either load the page in a
-  different browser or load the site from the HTTPS port:
+  different browser, e.g. Firefox, or load the site from the HTTPS port:
   https://myproject.ddev.site:8984
 * If https://myproject.ddev.site:8984 does not work make sure that the
   `HTTPS_EXPOSE=8984` line is present in the `docker-compose.solr.yml` file,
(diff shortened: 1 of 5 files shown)


docker-compose-services/solr-5/README.md
```diff
@@ -45,4 +45,59 @@ DDEV has a note that it will clean up volumes that match the service name. Since
...
+
+## Troubleshooting
+
+If the Solr service is not available to the client project (e.g. Drupal) using
+the expected URL format, the Solr index may not have been created correctly. It
+is possible to access Solr directly from the host OS by loading its full URL,
+e.g. `http://myproject.ddev.site:8983`.
+
+* Make note that the URL uses the HTTP protocol, not HTTPS.
+* Safari may insist in loading the URL as HTTPS, e.g,
```
rfay, line 85: Just add HTTPS_EXPOSE=8984 like the [core example has](https://github.com/drud/ddev/blob/master/pkg/servicetest/testdata/TestServices/docker-compose.solr.yaml) and then use `https://myproject.ddev.site:8984` everywhere and this problem goes away.

rfay (changes requested):
Might as well just solve the http/https redirection problem with HTTPS_EXPOSE.

docker-compose-services/solr-5/README.md
damienmckenna: Good call, thank you. I've updated the README files to mention this.

rfay:
I was saying just add the HTTPS_EXPOSE to the docker-compose.solr.yaml so you can leave that whole section out. Then change references to the web UI to use https.

rfay (approved):
Thanks for all the careful thinking here. Rebased and added a commit to add the HTTPS_EXPOSE to solr recipes that were missing it.

damienmckenna:
Thanks for the refinements, @rfay!

PR #149, opened 2021-08-16
Update README.md

siliconmeadow:
docker-compose.solr.yml needs to go into `.ddev` not `.ddev/solr`.

I couldn't use the recipe for solr4 until I'd realised the file was in the wrong place.

It updates  the instructions with the correct place to copy the file.


rfay (approved):
Thanks!

PR #152, opened 2021-09-19
Use official image for postgres

rfay:
The existing postgres recipe uses an unmaintained postgres image (`mdillon/postgis:11`), which also does not have arm64 images. 

This switches it to use the current official postgres image, `postgres:13.4`

@penyaskito @pcambra @rvolk @d3pendent @dacostafilipe Please make sure this is OK with you. I don't know why it was using an off-brand image to begin with.

### DIFF
--- a/docker-compose-services/postgres/docker-compose.postgres.yaml
+++ b/docker-compose-services/postgres/docker-compose.postgres.yaml
@@ -2,7 +2,7 @@ version: '3.6'
 services:
   postgres:
     container_name: ddev-${DDEV_SITENAME}-postgres
-    image: mdillon/postgis:11
+    image: postgres:13.4
     ports:
       - 32784:5432
     environment:


tyler36:
The docs will need to be updated for existing projects.

My current project uses `mdillon/postgis:11`. If I change this to `postgres:13.4`, I get the following message:

```
Failed to start test-psql: container failed to become healthy: err=container 'postgres' exited, please use 'ddev logs -s postgres' to find out why it failed 
```

- If I run the suggested command:
```
$ ddev logs -s postgres
2021-09-22 08:33:58.184 UTC [1] FATAL:  database files are incompatible with server
2021-09-22 08:33:58.184 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 11, which is not compatible with this version 13.4 (Debian 13.4-1.pgdg100+1).
```

## Fix

- The docker volume should be removed before updating:

```bash
docker stop 
docker volume rm ddev-test-psql_postgres
```

- This is destructive right? It destroys the database? But, the postgres backup too doesn't appear to be working (see #153 )
- Not sure if there is another way though.

dacostafilipe:
I used `mdillon/postgis` because I required the PostGIS plugin that was not part of the official PostgreSQL packages.

For my projects, a change to the official postgresql will not work as they require postgis, but maybe we should move this to a "postgis" recipe instead and rewrite this recipe to be a more plain "postgres" usage of the db?

@tyler36 Youn eed to export your data via native pg_dump (or pgsql_export as soon as it's fixed) before the migration to postgres 13

tyler36:
@dacostafilipe I guessed I was going to have to use pg_dump along the way.

I agree that keeping this recipe generic (via the official image), with additional PostGIS would be the better option. But I don't have a need for PostGIS. I actually need to downgrade to a postgres:10 image to match a project I'm on.

The official Postgres image [docs](https://hub.docker.com/_/postgres) says this regarding PostGIS:

```markdown
When using the default (Debian-based) variants, installing additional extensions (such as PostGIS) should be as simple as installing the relevant packages (see github.com/postgis/docker-postgis for a concrete example).
```

dacostafilipe:
@tyler36  Yes, we could also add a section describing how to add postgis via `dbimage_extra_packages` or `.ddev/db-build/Dockerfile`

dacostafilipe:
I tried a different way by using the `build` field on the docker-composer file instead of using an image directly:

```yaml
    build:
      context: ./postgis
      dockerfile: Dockerfile
```

With the same content as the official docker-postgis : https://github.com/postgis/docker-postgis/blob/master/13-3.1/Dockerfile

rfay:
I think using the official image is probably the best path, and perhaps a separate recipe for postgis. 

The build technique @dacostafilipe used should be perfect.

rfay:
I'm going to go ahead and pull this. Additions requested. I doubt that people with existing working postgres setups will be changing to latest docker-compose.postgres.yaml, and this will primarily be used on new projects.

PR #158, opened 2021-10-01
convert text to link directly to docs

tyler36:
This PR converts text to link directly to the referenced places in the documentation.

Because opening a new tab, typing search terms, scanning for the correct reference is to much effort.

- click text
- read docs


rfay (approved):
Works for me, thanks!

PR #160, opened 2021-10-01
Create additional databases for Postgres, fixes #159

### LINKED ISSUE #159: allow DDEV to add additional database configurations for Postres
investigate adding a second database on DDEV startup

tyler36:
I have a project that uses a postgres db. I usually use an im_memory sqlite database for testing because its fast. 

Recently, I was debugging an issue which came down to a difference between postgres & sqlite databases so I needed to change my tests to use postgres.  This started me down the path of "how do I add an additional db to a DDEV postgres container".

The offical way to add multiple databases is outlined on the docker image site under the "Initialization scripts".

We simply add a volume and it works. I've used the same folder as the `pgsql_export` command to reduce clutter.

This also adds several possibilies:
- Run any abritary query code.
- Remove the need to use hooks to import on first run
- If `pgsql_export` has previously run, it will auto-import it (cool)

This will *only* run if the volume is empty so it's non-destructive. 

Setup database:
- Setup a postgres database
- Add content
- Export with `ddev pgsql_export`.  Temporarily move this file somewhere else.

 Reset database:
- Power down ddev 
- Remove postgres volume for project
- Start DDEV and confirm database is empty

Making it work:
- Power down ddev 
- Move the previously created SQL file back to `./.ddev/psgql-db`
- Start DDEV & confirm database has restored.

### DIFF
--- a/docker-compose-services/postgres/README.md
+++ b/docker-compose-services/postgres/README.md
@@ -54,7 +54,7 @@ a `ddev pg_restore` command that will restore `.ddev/pgsql-db/postgresql.db.dump
 
 ## Multiple databases
 
-* By default, DDEV adds a single database called `db`. But it's possible to create additional databases when you first initialize this service.
+* By default, this recipe adds a single database called `db`. But it's possible to create additional databases when you first initialize this service.
 
 * The offical way to add multiple databases is outlined on the [docker image](https://hub.docker.com/_/postgres) under the "Initialization scripts"
 
@@ -81,22 +81,7 @@ EG. Lets create a second database called `testing`.
 
 * Note: All *.sql,*.sql.gz, & *.sh files in `./.ddev/pgsql-db` will be executed so you can create multiple database or even seed data.
 
-* Note: These files "are only run if you start the container with a data directory that is empty". IE. The first time you start the project. For exisiting projects, you must remove the volume first. Remember to backup the database first (Eg. `ddev pgsql_export`).
-
-### Removing a postgres volume
-
-  ```bash
-  # Display all docker volumes.
-  docker volume ls
-
-  # Stop DDEV.
-  ddev stop
-
-  # Remove 'example' project postgres volume.
-  docker volume rm ddev-example_postgres
-  ```
-
-* If you recieve a `Error response from daemon: ... volume is in use`, you need to stop DDEV first.
+* Note: These files "are only run if you start the container with a data directory that is empty". IE. The first time you start the project. For exisiting projects, add a hook to [import a sql dump if the database is empty](https://github.com/drud/ddev-contrib/tree/master/hook-examples/import-db-if-empty), or run `ddev delete` to re-initialize the project. Remember to backup the database first (Eg. `ddev pgsql_export`).
 
 ## PostGIS
 
--- a/docker-compose-services/postgres/README.md
+++ b/docker-compose-services/postgres/README.md
@@ -60,12 +60,11 @@ a `ddev pg_restore` command that will restore `.ddev/pgsql-db/postgresql.db.dump
 
 EG. Lets create a second database called `testing`.
 
-* Add an additional volume to `./.ddev/docker-composer.postgres.yaml`
+* Add `./pgsql-db:/docker-entrypoint-initdb.d` to the `volumes` section in to `./.ddev/docker-composer.postgres.yaml`
 
   ```yaml
       volumes:
-      ...
-      - "./pgsql-db:/docker-entrypoint-initdb.d"
+        - "./pgsql-db:/docker-entrypoint-initdb.d"
   ```
 
 * Create `./.ddev/pgsql-db` folder if it doesn't already exist
--- a/docker-compose-services/postgres/README.md
+++ b/docker-compose-services/postgres/README.md
@@ -99,11 +99,6 @@ To implement this in DDEV:
 
 * If you recieve a `Error response from daemon: ... volume is in use`, you need to stop DDEV first.
 
-### `ddev pgsql_export` auto-import
-
-* The file created by `ddev pgsql_export` is a great example of some of the commands you can use.
-* If this file exists, and the database data directory is empty (ie. you deleted the volume), the file will automatically by imported and populate your database!
-
 ## PostGIS
 
 The `postgres` image support `postgis`, but you will need to create the extension before using it:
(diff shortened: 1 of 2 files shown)


rfay:
@penyaskito @pcambra @rvolk @d3pendent @dacostafilipe ?

docker-compose-services/postgres/README.md
```diff
@@ -52,6 +52,58 @@ hooks:
...
+
+* The offical way to add multiple databases is outlined on the [docker image](https://hub.docker.com/_/postgres) under the "Initialization scripts"
+
+To implement this in DDEV:
+
+* Add an additional volume to `./.ddev/docker-composer.postgres.yaml`
+
+  ```yaml
+      volumes:
+      ...
```
pcambra, line 67: This might be confusing, better to put the example without it

docker-compose-services/postgres/README.md
```diff
@@ -52,6 +52,58 @@ hooks:
...
+* Create `./.ddev/pgsql-db` folder if it doesn't already exist
+
+  ```bash
+  mkdir ./.ddev/pgsql-db -p
+  ```
+
+* Add your *.sql,*.sql.gz, or *.sh files to `./.ddev/pgsql-db`. EG. Create a file called `./.ddev/pgsql-db/create_testing_database.sql` and add the following:
+
+  ```sql
+  CREATE DATABASE testing
```
pcambra, line 80: I think we should pick a name as mentioned before and run with it consistently in the example?

docker-compose-services/postgres/README.md
```diff
@@ -52,6 +52,58 @@ hooks:
...
+  # Remove 'example' project postgres volume.
+  docker volume rm ddev-example_postgres
+  ```
+
+* If you recieve a `Error response from daemon: ... volume is in use`, you need to stop DDEV first.
+
+### `ddev pgsql_export` auto-import
+
+* The file created by `ddev pgsql_export` is a great example of some of the commands you can use.
+* If this file exists, and the database data directory is empty (ie. you deleted the volume), the file will automatically by imported and populate your database!
```
pcambra, line 105: This is confusing, I think, pgsql_export will generate a db.sql that will be used for the primary "db", not sure how this relates

docker-compose-services/postgres/docker-compose.postgres.yaml
```diff
@@ -19,6 +19,7 @@ services:
         source: "."
         target: "/mnt/ddev_config"
       - ddev-global-cache:/mnt/ddev-global-cache
+      - "./pgsql-db:/docker-entrypoint-initdb.d"
```
pcambra, line 22: Is there a missing file on the PR?

pcambra (commented):
I've added some suggestions.

docker-compose-services/postgres/README.md
tyler36: Writing comments is not my strong point. English is hard.

docker-compose-services/postgres/docker-compose.postgres.yaml
tyler36: Not sure what you mean. 

- `./pgsql-db` is a local folder the docs mention to create. 
- `/docker-entrypoint-initdb.d`  is an endpoint used by the image.

Are you getting an error?

rfay:
Let me know when everybody thinks this is OK.

docker-compose-services/postgres/README.md
tyler36: Deleted

docker-compose-services/postgres/README.md
```diff
@@ -52,6 +52,52 @@ hooks:
 There are also another non-plain-text formats that `pg_dump` can generate, and you might need to work with them. If that's the case, there is also
 a `ddev pg_restore` command that will restore `.ddev/pgsql-db/postgresql.db.dump` into `db`.
 
+## Multiple databases
+
+* By default, DDEV adds a single database called `db`. But it's possible to create additional databases when you first initialize this service.
```
rfay, line 57:
```suggestion
* By default, this recipe adds a single database called `db`. But it's possible to create additional databases when you first initialize this service.
```

docker-compose-services/postgres/README.md
```diff
@@ -52,6 +52,52 @@ hooks:
...
+
+  ```bash
+  # Display all docker volumes.
+  docker volume ls
+
+  # Stop DDEV.
+  ddev stop
+
+  # Remove 'example' project postgres volume.
+  docker volume rm ddev-example_postgres
```
rfay, line 96: ddev automatically removes volumes that are properly named with `ddev delete`, so I don't think this set of instructions is needed.

docker-compose-services/postgres/README.md
tyler36: Good suggestion. I keep forgetting I'm not in the "supported" DDEV repo.

docker-compose-services/postgres/README.md
tyler36: You're right. `DDEV delete` is more elegant solution.  I feel nervous when "an existing project" and "delete" are used so close together.


Side note:
`DDEV delete` generates an error because it can't generate a snapshot.
```
Failed to remove project test-psql: 
unable to snapshot database, 
```

This project has the mysql database services disables it doesn't I don't need. Running `ddev stop --omit-snapshot` works, although I get a warning ...
```
Unable to delete drud/ddev-dbserver-mariadb-10.3:v1.18.0-test-psql-built: no such image
```
Not sure if DDEV should be trying to delete a service if it's been disabled or my config file still has some distant reference.

tyler36:
Personally, I prefer this "official" style volume solution over the use of DDEV hooks because:
- it transfer for other non-DDEV projects
- I can just dump a range of SQL-type files in the folder and it works
- its not just limited to creating a database

rfay:
Related:
* https://github.com/drud/ddev/issues/2187
I did prioritize that for the coming release cycle. Not sure if that was too optimistic.

rfay:
> Personally, I prefer this "official" style volume solution over the use of DDEV hooks because:
it transfer for other non-DDEV projects
I can just dump a range of SQL-type files in the folder and it works
its not just limited to creating a database

But you removed that right?

As far as killing off databases (if you add that back) 
* It would be good to actually adapt https://github.com/drud/ddev-contrib/tree/master/hook-examples/import-db-if-empty to pgsql
* Why not consider [pg_drop](https://github.com/drud/ddev-contrib/blob/master/docker-compose-services/postgres/commands/postgres/pgsql_drop) as the way to delete the databases, instead of introducing people to arcane docker things? (or just `ddev exec -s postgres dropdb -U db <dbname>`).  There's also a `--if-exists` and `--force` argument.

With an approach like that, people would be using the postgres commands they're familiar with instead of trying to understand why they'd use a docker volume to delete a database.

tyler36:
I can't seem to see any open requested changes for this? 

How would you like to proceed?

rfay:
What do you think about this now @pcambra  ? Thanks!

PR #161, opened 2021-10-01
update pgsql_export command to use stated directory

penance316:
The description and documentation mentions the PG export will go into location like '.ddev/pgsql-db/postgresql.db.sql'. 
This update ensures that to be true.

### DIFF
--- a/docker-compose-services/postgres/commands/postgres/pgsql_export
+++ b/docker-compose-services/postgres/commands/postgres/pgsql_export
@@ -4,4 +4,4 @@
 ## Usage: pgsql_export
 ## Example: ddev pgsql_export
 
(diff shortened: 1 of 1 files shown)


rfay:
@tyler36 correct?

penance316:
These are the lines in documentation I was referring to. 🤔 
https://github.com/drud/ddev-contrib/blob/ff8e437d539816aa257ee3a0509757cf1975386a/docker-compose-services/postgres/README.md#L35

https://github.com/drud/ddev-contrib/blob/db26514f3449c81df791c2d89d0f96ada478e785/docker-compose-services/postgres/commands/postgres/pgsql_export#L3

tyler36:
good catch!

PR #177, opened 2021-11-24
Fixed ambiguous RabbitMQ host

felmab:
Same issue as in https://stackoverflow.com/a/69943996/17395654


rfay:
Already fixed in https://github.com/drud/ddev-contrib/pull/176 right?

felmab:
Oops, that's right, I should have checked this before. I also made changes in the Configuration section, though.

rfay:
Will take this one then and fix the other one.

PR #178, opened 2021-11-24
Fixed port and hostname formatting

felmab:
- Default AMQP port is always 5672 and not 15672, whether from the host or a container.
- Formatting issue with hostname in configuration section.


rfay:
Thanks so much for maintaining this!

PR #179, opened 2021-12-06
Add command to stop all projects except current

stborchert:
The idea came up on Discord by @sceo (see https://discord.com/channels/664580571770388500/689166305315520541/915724577261359104).

Original question:

> Is there a way to stop all running ddev projects except the one you're currently in?

@rfay came up with a handy one-liner utilizing `jq` that is added as a custom command here.

### DIFF
--- a/custom-commands/stop-other/README.md
+++ b/custom-commands/stop-other/README.md
@@ -4,5 +4,7 @@ Copy the [stop-other](./stop-other) command to the .ddev/commands/host/ director
 
 Now you can run `ddev stop-other` to stop all running projects except the one you are calling the command from.
 
+This requires that the `jq` command be installed. `brew install jq` or see https://stedolan.github.io/jq/download/
+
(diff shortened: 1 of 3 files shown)


custom-commands/stop-other/README.md
```diff
@@ -0,0 +1,8 @@
+# Stops all running projects except the current.
+
+Copy the [stop-other](./stop-other) command to the .ddev/commands/host/ directory or the global ~/.ddev/commands/host directory (to make it work for all projects).
+
+Now you can run `ddev stop-other` to stop all running projects except the one you are calling the command from.
```
rfay, line 5:
```suggestion
Now you can run `ddev stop-other` to stop all running projects except the one you are calling the command from.

This requires that the `jq` command be installed. `brew install jq` or see https://stedolan.github.io/jq/download/
```

rfay (changes requested):
Needs an entry in top-level readme, thanks.

rfay (approved):
Thanks!

PR #156, opened 2021-09-27
Rename directory for pgsql import/export from import-db to psql-db

tyler36:
## The New Solution/Problem/Issue/Bug:
This is perhaps a matter of personal preference, but I think renaming `import-db` to a more specific `pgsql-db` makes it's purpose more clear. 

## How this PR Solves The Problem:

- The folder is used by the pgsql command
- The folder is specific for postgres
- The folder is not just for importing, but general "db" matters

The documents refer to:
- postgresql
- postgres
- postgre (although this has one reference and might be a typo)
- pgsql
- psql

`pgsql-db` was chosen as the new folder name to link the commands (pgsql_export, pgsql_import etc.) with it's purpose (database).

I'm happy for the name to be changed. This felt like the simpliest and shortest at the time.

PR #184, opened 2021-12-10
PostgreSQL commands: changed from `pgsql-db` directory to `import-db`

### LINKED ISSUE #183: PostgreSQL commands not working (directory permissions)
Hi all, 

I noticed that the PostgreSQL commands (https://github.com/drud/ddev-contrib/tree/master/docker-compose-services/postgres/commands/postgres) do not work out of the box.

I was able to find the reason relatively quickly: In the comments the export & import directory ".ddev/pgsql-db" is mentioned, this is also set correctly during pg_dump and import, but the pgsql_export script still says that the directory "/mnt/ddev_config/import-db" and not "pgsql-db" should be created.

When I changed this to "pgsql-db" everything worked as well. However, the "pgsql-db" folder is not included in .gitignore, so I wonder if it doesn't make sense again to store the database dump in "import-db".

tschni:
This PR fixes #183 by running the PostgreSQL export & import process entirely through the `import-db` directory.

The `import-db` has the advantage over the `pgsql-db` that it is already entered in the .gitignore file.

### DIFF
--- a/docker-compose-services/postgres/README.md
+++ b/docker-compose-services/postgres/README.md
@@ -32,8 +32,8 @@ When using multiple project with PostgreSQL support, remember to update your `do
 
 Two new `ddev` commands are provided:
 
-* `ddev pgsql_export` : Use `pg_dump` to export `db` to `.ddev/pgsql-db/postgresql.db.sql`
-* `ddev pgsql_import` : Use `pgsql` to import `.ddev/pgsql-db/postgresql.db.sql` into `db` - Note that this must be executed with an empty database.
+* `ddev pgsql_export` : Use `pg_dump` to export `db` to `.ddev/import-db/postgresql.db.sql`
+* `ddev pgsql_import` : Use `pgsql` to import `.ddev/import-db/postgresql.db.sql` into `db` - Note that this must be executed with an empty database.
 
 Example `config.yaml` hooks configuration to automatically import/export the `db` table:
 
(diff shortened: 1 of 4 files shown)


tyler36:
This PR fixes the #183  issue, but does so by reverting #156  (my PR).

My original PR was (and the repo still is) missing the `mkdir`, `chmod` changes required. I'm amazed I hadn't noticed it.

```diff
- mkdir -p /mnt/ddev_config/import-db && chmod 777 /mnt/ddev_config/import-db && su postgres -c "pg_dump db --username=db --host=localhost --port=5432 --column-inserts > /mnt/ddev_config/pgsql-db/postgresql.db.sql"
+ mkdir -p /mnt/ddev_config/pgsql-db && chmod 777 /mnt/ddev_config/pgsql-db && su postgres -c "pg_dump db --username=db --host=localhost --port=5432 --column-inserts > /mnt/ddev_config/pgsql-db/postgresql.db.sql"
```


Now, we need to decide how to handle DDEV import/exports from different database implementations:
1. Everything goes in a `import-db` folder. (this PR)
2. Split postgreSQL into a seperate folder (`pgsql-db`), keeping the current MySQL-like DBs untouched.
3. Create a folder structure:
  -  `./.ddev/db/mysql`
  - `./.ddev/db/postgres`
  
I'm not sure how many people are running multiple database types in a project. 
I'm not sure if DDEV is/will expand to include other database types.
I'm not a fan of `import-db` mostly because it doesn't feel like an intuitive place where the database is exported ( or other DB services); a generic name like `db` seem better to me, or a folder for the service (like my original PR).

A decision here should also affect how we want to "standardize" services / addons for DDEV and DDEV-contrib.

tschni:
Thanks for your comment. What bothered me about the `pgsql-db` directory is that it was not automatically in the `.gitignore` file.

Since the file has a unique name (`postgresql.db.sql`), the `import-db` directory should not really be a problem. But I prefer it just because of the .gitignore file.

rfay (approved):

PR #185, opened 2021-12-13
Add an initial Colima guide

deviantintegral:

It's time to document what we've learned at https://github.com/drud/ddev/issues/3208

This PR adds a "from scratch" guide for using colima instead of Docker Desktop.

Follow the instructions in this doc!

### DIFF
--- a/recipes/colima/README.md
+++ b/recipes/colima/README.md
@@ -1,0 +1,30 @@
+# Using Colima instead of Docker Desktop on macOS
+
+[Colima](https://github.com/abiosoft/colima) is an alternative runtime to Docker Desktop with a permissive license. Colima is built on top of [Lima](https://github.com/lima-vm/lima) which is a more general virtual machine application.
+
+Reasons to use Colima include:
+
+- Preferring to use open-source software (Docker Desktop, unlike Docker, is proprietary software).
+- Working for an organization that due to its size requires a paid Docker plan to use Docker Desktop, and wanting to avoid that cost and business relationship.
+- Preferring a CLI-focused approach to Docker Desktop's GUI focus.
+
+In day-to-day use, Colima is very similar to how WSL2 works on Windows.
+
+## Using Colima with DDEV
+
+1. Follow the [instructions to install Colima](https://github.com/abiosoft/colima#installation). Colima is in homebrew which is also a requirement for ddev.
+1. If not already installed, install the Docker command line tools on macOS with `brew install docker`.
+1. Run `colima start` to create the VM. For typical DDEV projects, it is recommended to use `colima start --cpu 4 --memory 4 --disk 100` as minimums to allow for 4 virtual CPUs, 4GB of memory, and 100GB of disk space.
+  - This first boot will take some time, as it will download an Ubuntu disk image, and then install docker within it. Generally this is only done once.
+1. If Docker Desktop is installed or running, by default command line tools will connect to it instead of Colima. Running `docker context use colima` will switch the active connection. Use `docker context ls` to see all available contexts.
+1. Run `docker version` to verify everything is working correctly. Both client and server information should be shown.
+1. Run `ddev config global --mutagen-enabled=true` to enable Mutagen syncing of code. By default, colima mounts are **read only** and very slow.
+1. Run `ddev start` like normal in the projects of your choice.
+1. Run `colima stop` to completely shut down the VM. This is equivalent to quitting Docker Desktop.
+- When starting your day, `colima start && ddev start` will start colima and your project all at once.
+- There is no need to run `ddev stop` before running `colima stop`.
+
+## Alternative configurations
+
+- Instead of using Mutagen, colima can be started with the `--mount` flag, such as with `--mount $HOME/projects:w` to mount the projects folder in your home directory read-write. However, colima currently uses sshfs for mounts and this is not suitable for many or most PHP or Node projects.
+- DDEV can also be installed _inside_ the virtual machine. This is identical to how WSL2 installations are managed. A notable limitation is that ddev cannot install TLS certificates from `mkcert` in macOS browsers. Consider this setup experimental, such as for testing out PHPStorm's new remote develompment features.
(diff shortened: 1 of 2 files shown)


deviantintegral:
I was going to document xdebug and `host.docker.internal`, but I need to followup on https://github.com/lima-vm/lima/pull/457/files first.

rfay:
It's great, but I'm working on https://github.com/drud/ddev/pull/3445 which will make import-db, export-db and hopefully snapshot features work as well. And I wouldn't be surprised if we can just get this into an experimental topic in ddev itself. Also, you can make lots of things work across DOCKER_HOST with this technique, colima is just one example.

deviantintegral:
Well, this is weird. I just rebooted for the macOS 12.1, and while docker at the CLI works, ddev throws:

```
$ ddev status
Failed to describe project(s): Get "http://unix.sock/containers/json?all=1&filters=%7B%22label%22%3A%5B%22com.ddev.site-name%3Ddrupal%22%2C%22com.docker.compose.service%3Dweb%22%5D%7D": dial unix /var/run/docker.sock: connect: no such file or directory
```

I actually uninstalled Docker Desktop earlier, so it's no longer in play. After running `colima start` I see:

```
$ docker context ls
NAME       DESCRIPTION                               DOCKER ENDPOINT                            KUBERNETES ENDPOINT   ORCHESTRATOR
colima *   colima                                    unix:///Users/andrew/.colima/docker.sock
default    Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                                      swarm
```

And if I symlink `/Users/andrew/.colima/docker.sock` to `/var/run` (using sudo, the directory isn't writable by a normal macOS user) `ddev` works.

So... perhaps I symlinked a month or two ago and forgot? I hadn't rebooted for a while. But, I would have thought that would have been brought up early in your testing of colima too.

rfay:
I also just upgraded to macOS 12.1 (moments ago). I'm not seeing general problems. Are you saying you're having problems just with colima, or in general with Docker Desktop?

I just exited Docker Desktop and started colima and had no trouble with it. However, I'm using the default docker context, I hadn't used `docker context` to switch back and forth with colima previously, will look forward to learning about that technique.

rfay:
BTW, one of the awesome things about colima is that we'll be able to use Github Actions to test it, even on macOS I think, since we don't have the crazy problem of how to install Docker Desktop.

recipes/colima/README.md
```diff
@@ -0,0 +1,32 @@
...
+1. Run `docker version` to verify everything is working correctly. Both client and server information should be shown.
+1. Run `ddev config global --mutagen-enabled=true` to enable Mutagen syncing of code. By default, colima mounts are **read only** and very slow.
+1. Run `ddev start` like normal in the projects of your choice.
+1. Run `colima stop` to completely shut down the VM. This is equivalent to quitting Docker Desktop.
+- When starting your day, `colima start && ddev start` will start colima and your project all at once.
+- There is no need to run `ddev stop` before running `colima stop`.
+
+## Alternative configurations
+
+- Instead of using Mutagen, colima can be started with the `--mount` flag, such as with `--mount $HOME/projects:w` to mount the projects folder in your home directory read-write. However, colima currently uses sshfs for mounts and this is not suitable for many or most PHP or Node projects.
```
abiosoft, line 29: `$HOME` is actually now mounted read-write by default and this step is no longer required by the user.
However, the concern for volume speed remains until it is fixed upstream in Lima.

deviantintegral:
It looks like ddev is broken in colima HEAD, soon to be 0.3.0.

https://github.com/abiosoft/colima/pull/40 switched to using docker contexts instead of hardcoding the socket, which I think is a good improvement. However, I guess I didn't realize that ddev doesn't actually use contexts itself.

As a workaround, we can manually add the symlink to the socket. However, I think it would be smoother if ddev respected the active docker context.

@rfay would you like me to update the docs with that workaround so we can merge as is, or would you rather leave this open, add context support in ddev, and merge this after?

rfay:
Shoot, the problem is that neither docker's go api nor docker-go respect contexts. Glad you caught this. I'll have to take a look at what the options are for respecting contexts. It's not likely to be pretty, but should be do-able.

I'd prefer to keep this open for now, please feel free to push new edits into it.

deviantintegral:
`docker context inspect` at least gives JSON output of the currently active context. No idea if that's considered to be a stable API though.

```
$ docker context inspect

[
    {
        "Name": "colima",
        "Metadata": {
            "Description": "colima"
        },
        "Endpoints": {
            "docker": {
                "Host": "unix:///Users/andrew/.colima/docker.sock",
                "SkipTLSVerify": false
            }
        },
        "TLSMaterial": {},
        "Storage": {
            "MetadataPath": "/Users/andrew/.docker/contexts/meta/f24fd3749c1368328e2b149bec149cb6795619f244c5b584e844961215dadd16",
            "TLSPath": "/Users/andrew/.docker/contexts/tls/f24fd3749c1368328e2b149bec149cb6795619f244c5b584e844961215dadd16"
        }
    }
]
```

rfay:
Yeah, the problem is that contexts are a level higher than the docker api, which assumes it knows who it's talking to.  I'll take a look at the docker cli code to see how it does it.

abiosoft:
> Shoot, the problem is that neither docker's go api nor docker-go respect contexts. Glad you caught this. I'll have to take a look at what the options are for respecting contexts. It's not likely to be pretty, but should be do-able.
> 

As a last resort, there is the `DOCKER_HOST` environment variable.

rfay:
Right, I'm pretty sure DOCKER_HOST would work right now.

rfay:
* https://github.com/drud/ddev/pull/3452 addresses this in a fairly sad way.

deviantintegral:
Nice! It's better than not working all.

Unfortunately, my laptop just bit the dust and it's 7-10 days to repair. I'm stuck on Windows in the meantime, so I won't be able to pick up any testing until January.

rfay:
Thanks for all your work on this! You should be able to test it in wsl2 though :)

rfay:
It may be that just using DOCKER_HOST will be a better answer than this PR. I'll have to take a look at the test failures and see how deep I got myself in trouble.

rfay:
Colima is now officially supported in HEAD, `brew unlink ddev && brew install drud/ddev/ddev --HEAD`, see https://ddev.readthedocs.io/en/latest/users/docker_installation/#macos-installation-colima

Thank you so much for pioneering this, and looking forward to your experiences, feedback, improvements.

PR #186, opened 2022-01-03
Service meilisearch v0.20.0

thilohille:
This receipe adds a meilisearch container to the project.
Meilisearch is an open source search-engine and can be used to handle searchindexes for the Drupal search_api.

Follow the instructions from docker-compose-services/meilisearch/README.md.
Expected result:
* a working meilisearch instance is available in the project 
* the meilisearch backend is reachable at http://<projectname>.ddev.site:7700

### DIFF
--- a/docker-compose-services/meilisearch/README.md
+++ b/docker-compose-services/meilisearch/README.md
@@ -1,0 +1,22 @@
+# Meilisearch
+
+Using official meilisearch v0.20.0 container [meilisearch](https://hub.docker.com/r/getmeili/meilisearch).
+
+## Installation
+
+Copy [docker-compose.meilisearch.yaml](docker-compose.meilisearch.yaml) to your project's .ddev folder.
+
+## Configuration
+
+From within the container, the meilisearch container is reached at hostname: ddev-<projectname>-meilisearch, port: 7700, so the server URL might be `http://ddev-<projectname>-meilisearch:7700`.
+
+## Connection
+
+You can access the Meilisearch server directly from the host for debugging purposes by visiting `http://<projectname>.ddev.site:7700`.
+
+You can use `ddev logs -s meilisearch` to investigate what the meilissearch daemon has been up to.
+
+## Additional Resources
+
+* https://www.drupal.org/project/search_api_meilisearch
+* https://docs.meilisearch.com/learn/getting_started/quick_start.html
--- a/docker-compose-services/meilisearch/README.md
+++ b/docker-compose-services/meilisearch/README.md
@@ -8,7 +8,7 @@ Copy [docker-compose.meilisearch.yaml](docker-compose.meilisearch.yaml) to your 
 
 ## Configuration
 
-From within the container, the meilisearch container is reached at hostname: ddev-<projectname>-meilisearch, port: 7700, so the server URL might be `http://ddev-<projectname>-meilisearch:7700`.
+From within the container, the meilisearch container is reached at hostname: ddev-\<projectname\>-meilisearch, port: 7700, so the server URL might be `http://ddev-<projectname>-meilisearch:7700`.
 
 ## Connection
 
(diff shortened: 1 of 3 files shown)


docker-compose-services/meilisearch/docker-compose.meilisearch.yaml
```diff
@@ -0,0 +1,27 @@
+version: '0.36'
+services:
+  meilisearch:
+    container_name: ddev-${DDEV_SITENAME}-meilisearch
+    image: getmeili/meilisearch:v0.20.0
+    hostname: ${DDEV_SITENAME}-meilisearch
+    ports:
+      - 7700:7700
```
rfay, line 8: Please use `expose` instead of ports - otherwise only one of these projects can be running at a time.

```suggestion
    expose:
      - 7700
```

docker-compose-services/meilisearch/README.md
```diff
@@ -0,0 +1,22 @@
...
+## Connection
+
+You can access the Meilisearch server directly from the host for debugging purposes by visiting `http://<projectname>.ddev.site:7700`.
+
+You can use `ddev logs -s meilisearch` to investigate what the meilissearch daemon has been up to.
+
+## Additional Resources
+
+* https://www.drupal.org/project/search_api_meilisearch
+* https://docs.meilisearch.com/learn/getting_started/quick_start.html
```
rfay, line 22: Please add "Contributed by" footer as in many other services.

rfay (changes requested):
* Please add a link in the top-level readme
* Please explain what meilisearch is in the README here


Thanks for this contribution!

docker-compose-services/meilisearch/docker-compose.meilisearch.yaml
```diff
@@ -0,0 +1,27 @@
+version: '0.36'
```
luenemam, line 1: Should be something like `version: '3.6'`

rfay:
```suggestion
version: '3.6'
```

docker-compose-services/meilisearch/docker-compose.meilisearch.yaml
thilohille: thanks for the quick review! the canfiguration has been changed to use "expose"

docker-compose-services/meilisearch/README.md
thilohille: done

docker-compose-services/meilisearch/docker-compose.meilisearch.yaml
thilohille: has been changed to '3.6'. thanks for pointing out.

rfay:
Still needs an entry in the top-level README.

thilohille:
> Still needs an entry in the top-level README.
Done

docker-compose-services/meilisearch/docker-compose.meilisearch.yaml
```diff
@@ -0,0 +1,30 @@
+version: '3.6'
+services:
+  meilisearch:
+    container_name: ddev-${DDEV_SITENAME}-meilisearch
+    image: getmeili/meilisearch:v0.20.0
```
mgralikowski, line 5: Why so old version? We have v0.0.25 already.

luenemam: We need it for a Drupal integration module [search_api_meilisearch](https://www.drupal.org/project/search_api_meilisearch). The maintainers declare compatibility with meilisearch:0.20.0.

rfay (approved):
I made a couple of changes to the README. most importantly, this is an ancient version of the meilisearch image, and it does not have arm64 support, so all those folks with the new M1s can't use it. I stumbled on that.

I went ahead and tried to test it on gitpod.io, but wasn't confident of what I was doing, but seemed it worked.

Please do lobby the maintainer of search_api_meilisearch to use a recent version, so a recent version can be used from dockerhub. It's always possible that the maintainer already has support, as they say there's been ongoing dev in https://www.drupal.org/project/search_api_meilisearch/issues/3221742... but it doesn't look to me like there has in fact been anything going on with this module. (no commits for a year).

PR #188, opened 2022-01-11
Prepare "networks:" and "links:" for ddev v1.19

jonaseberle:
For ddev v1.19 we'll have to update services that need to be reachable or reach the router or inter-project services.

This removes superfluous `links:` and adds `networks: [ default, ddev ]` were inter-project communication is necessary.

## Related Issue Link(s):
https://github.com/drud/ddev/pull/3403

### DIFF
--- a/docker-compose-services/drupalci-chromedriver/docker-compose.chromedriver.yaml
+++ b/docker-compose-services/drupalci-chromedriver/docker-compose.chromedriver.yaml
@@ -7,16 +7,15 @@ version: '3.6'
 services:
   chromedriver:
     image:  drupalci/webdriver-chromedriver:production
+    networks: [default, ddev_default]
     container_name: ddev-${DDEV_SITENAME}-chromedriver
     labels:
       com.ddev.site-name: ${DDEV_SITENAME}
       com.ddev.approot: $DDEV_APPROOT
     external_links:
-      - ddev-router:${DDEV_SITENAME}.${DDEV_TLD}
+      - ddev-router:${DDEV_HOSTNAME}
 
   web:
-    links:
-      - chromedriver
     environment:
       # *** One of these must be uncommented ***
       # In order for the system to work, one of these must be uncommented so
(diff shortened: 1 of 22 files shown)


rfay (changes requested):
Thanks for being so proactive about this! 

People are already having trouble with the documented [solr](https://github.com/drud/ddev/blob/master/pkg/servicetest/testdata/TestServices/docker-compose.solr.yaml), both people who are still on v1.18 and people who are trying out the v1.19.0-alpha1. We need to figure out a way to help people with both of those.

The people who are using solr for the first time hit the networks issue when using v1.18.2 because "ddev" doesn't exist. They can work around this with `docker network create ddev` but that will have to be mentioned somehow. Note that the docs point to the latest version of the docker-compose.solr.yaml, which doesn't actually work as-is with v1.18.2.

The people who already have solr and upgrade to v1.19.0-alpha1 have trouble because they're missing the networks stanza. But the trouble is quite mysterious - the ddev-router fails to come up because it can't find resolve the hostname. ddev-router gets notification of the new container coming up but can't resolve or contact it. 

I'll open a PR to try to help this on solr in a few minutes, will appreciate your thoughts and review. I did already add a caveat in the [release notes](https://github.com/drud/ddev/releases/tag/v1.19.0-alpha1), but it probably needs more work also.

jonaseberle:
This was not meant to be pulled before ddev 1.19 is out.

Oh, is https://github.com/drud/ddev/blob/master/pkg/servicetest/testdata/TestServices/docker-compose.solr.yaml linked somewhere in the docs? I think we could say this is not meant to be used before 1.19.

I also thought about how to make ddev projects "ready" for v1.19 without breaking v1.18 but have not come up with a good idea. 

I am back to work since today and I'll monitor discord/Slack now again and try to support. Also watching ddev issues/pulls.

rfay (changes requested):
Needs to change from `networks: [default, ddev]` to `networks: [default, ddev_default]`

jonaseberle:
rebased and added `networks: [default, ddev_default]` to ALL services.

rfay:
It may still be a month or more before ready, but thanks for maintaining this! I still hope to get support for postgres into this release.

rfay:
We don't have to do this any more since ddev now adds the networks stanza, closing.

rfay:
And thanks for *all* your work on this!

PR #189, opened 2022-01-14
Add login command to custom commands 

bserem:
## Custom command to automatically login to Drupal

Provide a `ddev login` command that automatically opens the default browser and logs in to the current Drupal project.


rfay (changes requested):
I think people will like this! 
* Please rename to drupal-login or something, since ddev isn't a drupal-specific tool. You can keep the name of the command as "login", but the directory name here and title and such should mention drupal.
* Please add an entry to the top-level README.md

rfay (approved):
Worked nicely for me on macOS. didn't test elsewhere. Thanks!

It would be a little better if it checked for drush, but that's slow on a host-side custom command.

bserem:
I tested it on Linux, I'll ask for a test on windows too.

Thanks

dimitriskr:
Hello

`ddev login` works great in WSL

PR #190, opened 2022-01-17
Fix documentation for getting old PHP versions working

darrenoh:
I had to do a little more than the README said to get Drupal working with an old version of PHP on ddev.

### DIFF
--- a/docker-compose-services/old_php/README.md
+++ b/docker-compose-services/old_php/README.md
@@ -1,14 +1,14 @@
-# Using Old PHP Versions (pre-php5.6)
+# Using old PHP versions (pre-php5.6)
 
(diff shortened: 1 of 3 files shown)


docker-compose-services/old_php/README.md
```diff
@@ -54,6 +54,9 @@ Now, if you're paranoid, do a `ddev restart` (probably unnecessary) and begin to
 
 The site you've set up can probably be used as a migration source, can be archived, converted to html with a sitesucker, etc.
 
+## Configure PHP settings
```
rfay, line 57: Maybe add a demonstration `.user.ini` file?

rfay (changes requested):
Thanks so much for chasing this all down! I didn't manually try it, but we did it the other day. 

I guess a couple of requests - 

1. Demonstrate usage of .user.ini, since I'm not familiar with it. Give a link to usage.
2. Is .user.ini only for Apache? If so, mention that.
3. Please put your name on the bottom as a maintainer of this, since you're now the local expert.  Change to "Contributed by @darrenoh and @rfay". Thanks!

rfay (approved):
Awesome, thanks!

PR #194, opened 2022-01-20
add Drupal config for redis [skip ci]

tyler36:
Expand redis docs with Drupal quickstart

### DIFF
--- a/docker-compose-services/redis/README.md
+++ b/docker-compose-services/redis/README.md
@@ -49,9 +49,6 @@ ddev redis-cli
 
 * Run `ddev describe` to check the status and ports used by the service.
 
-
-## Framework Quickstarts
-
 * Frameworks require configuration to use the redis service.
 * You need to replace `<DDEV_SITENAME>` values with your project's site name. Hint: `ddev describe`
 
@@ -63,12 +60,16 @@ ddev redis-cli
 ddev composer require drupal/redis
 ```
 
-* add setting information to `web/sites/default/settings.local.php`; remember to update '<DDEV_SITENAME>'
+* add the service to `web/sites/default/settings.php`; a great place is inside the ddev block.
 
 ```php
-# DDEV SERVICE: Redis
-$settings['redis.connection']['host']      = 'ddev-${DDEV_SITENAME}-redis';  // Your Redis instance hostname.
-$settings['cache']['default'] = 'cache.backend.redis';
+if (file_exists(__DIR__ . '/settings.ddev.php') && getenv('IS_DDEV_PROJECT') == 'true') {
+  include __DIR__ . '/settings.ddev.php';
+
+  # DDEV SERVICE: Redis
+  $settings['redis.connection']['host'] = getenv('DDEV_SITENAME');
+  $settings['cache']['default'] = 'cache.backend.redis';
+}
 ```
 
 * clear Drupal caches with `drush`
--- a/docker-compose-services/redis/README.md
+++ b/docker-compose-services/redis/README.md
@@ -49,16 +49,6 @@ ddev redis-cli
 
 * Run `ddev describe` to check the status and ports used by the service.
 
-```shell
-$ ddev describe
-...
-├──────────┬──────┬─────────────────────────────────────────┬───────────────────────┤
-│ SERVICE  │ STAT │ URL/PORT                                │ INFO                  │
-├──────────┼──────┼─────────────────────────────────────────┼───────────────────────┤
-│ redis    │ OK   │ https://example.ddev.site               │                       │
-│          │      │ InDocker: ddev-example-redis:6379       │                       │
-│          │      │ Host: localhost:54710                   │                       │
-```
 
 ## Framework Quickstarts
 
@@ -73,19 +63,11 @@ $ ddev describe
 ddev composer require drupal/redis
 ```
 
-* enable the module via the website (`/admin/modules`) or via `drush` with the following command
-
-```shell
-$ ddev drush en redis
- [success] Successfully enabled: redis
-```
-
-* add setting information to `web/sites/default/settings.ddev.php`; remember to update '<DDEV_SITENAME>'
+* add setting information to `web/sites/default/settings.local.php`; remember to update '<DDEV_SITENAME>'
 
 ```php
 # DDEV SERVICE: Redis
-$settings['redis.connection']['interface'] = 'PhpRedis'; // Can be "Predis".
-$settings['redis.connection']['host']      = '<DDEV_SITENAME>';  // Your Redis instance hostname.
+$settings['redis.connection']['host']      = 'ddev-${DDEV_SITENAME}-redis';  // Your Redis instance hostname.
 $settings['cache']['default'] = 'cache.backend.redis';
 ```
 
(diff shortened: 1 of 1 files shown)


rfay:
@gormus @tommym9 Could you please take a look at this PR? Or if you know of some other knowledgable reviewers @tyler36 .

docker-compose-services/redis/README.md
```diff
@@ -60,27 +60,70 @@ $ ddev describe
...
+This example shows:
 
-Frameworks, such as Laravel, require configuration to use the redis service.
+* From the webserver point-of-view:
+  * REDIS HOST: ddev-example-redis
+  * REDIS PORT: 6379
+
+* From the host Operating System point-of-view:
+  * REDIS HOST: localhost
+  * REDIS PORT: 54710
```
gormus, line 71: The `docker-compose.redis.yml` file should have a reference to internal port number `6379`, yes.

But the port exposed to host machine will likely to differ for each user depending on their Docker usages. 

I don't think it's a good idea to include to port for the host machine without any explanation. Please consider adding an explanation that port number will likely to be different for the user.

docker-compose-services/redis/README.md
```diff
@@ -60,27 +60,70 @@ $ ddev describe
...
+$ ddev drush en redis
+ [success] Successfully enabled: redis
+```
+
+* add setting information to `web/sites/default/settings.ddev.php`; remember to update `<REDIS_HOST>`
+
+```php
+# DDEV SERVICE: Redis
+$settings['redis.connection']['interface'] = 'PhpRedis'; // Can be "Predis".
+$settings['redis.connection']['host']      = '<REDIS_HOST>';  // Your Redis instance hostname.
```
gormus, line 98: You could use `redis` as the Redis service hostname as defined in here: https://github.com/drud/ddev-contrib/blob/master/docker-compose-services/redis/docker-compose.redis.yaml#L6 

```php
$settings['redis.connection']['host'] = 'redis';
$settings['redis.connection']['port'] = 6379;
```

same also on lines 114, and 118

docker-compose-services/redis/README.md
```diff
@@ -60,27 +60,70 @@ $ ddev describe
 │          │      │ Host: localhost:54710                   │                       │
 ```
 
-## Examples
+This example shows:
 
-Frameworks, such as Laravel, require configuration to use the redis service.
+* From the webserver point-of-view:
+  * REDIS HOST: ddev-example-redis
```
gormus, line 66: May I suggest to replace this with `ddev-${DDEV_SITENAME}-redis` instead? 

...to emphasize what part of the name is dynamically defined.

gormus (commented):
I think some further clarification needed.

I am also hesitant to include instructions for setting up the apps. 

Specifically for Drupal, there are various options for configuring the Redis module. And, not all options are suitable for all projects.

Perhaps, including an example for a bare-minimum config, and another for an hypothetical complex setup could provide developers more insight, depending on their experiences.

tyler36:
> Specifically for Drupal, there are various options for configuring the Redis module. And, not all options are suitable for all projects.

Yes, but this can be said for any service.

> Perhaps, including an example for a bare-minimum config, and another for an hypothetical complex setup could provide developers more insight, depending on their experiences.

I think this does include the bare minimum config for Durpa. I think adding a complex setup would defeat the purpose of quickstart.

docker-compose-services/redis/README.md
tyler36: The idea was to provide an example output and teach users how to read it.  The URL and hostname are also project specific but I think people will be able to follow.

docker-compose-services/redis/README.md
```diff
@@ -60,27 +60,70 @@ $ ddev describe
...
+
+## Drupal
+
+* install the [`redis` module](https://www.drupal.org/project/redis)
+
+```shell
+ddev composer require drupal/redis
+```
+
+* enable the module via the website (`/admin/modules`) or via `drush` with the following command
```
tommym9, line 86: So you don't *need* to enable the Redis module. This will work happily without it. I would say turning it on enables you to see the connection in the status report and a detailed view at `/admin/reports/redis`

tommym9:
I've got a bit of time this afternoon (GMT) to go through this from scratch on a fresh Drupal site. Will give it a go with the above instructions and report back.

tommym9:
So turns out I have this on a site I'm working on already. The only config I have for Drupal is:

```
$settings['cache']['default'] = 'cache.backend.redis';
$settings['redis.connection']['host'] = 'ddev-{host}-redis';
$settings['redis.connection']['port'] = 6379;
```

In my `settings.local.php`. This is included by `settings.php` and doesn't go into git.

The docker compose file is the base one in additional services repo: https://github.com/drud/ddev-contrib/tree/master/docker-compose-services/redis

I had to pull in the  `predis/predis` package, but I didn't need to specify it as the interface as it looks like its the default.

This is a Drupal 9 site for reference.

Cheers

Tom

docker-compose-services/redis/README.md
```diff
@@ -49,20 +49,35 @@ ddev redis-cli
...
-│ redis    │ OK   │ https://example.ddev.site               │                       │
-│          │      │ InDocker: ddev-example-redis:6379       │                       │
-│          │      │ Host: localhost:54710                   │                       │
+ddev composer require drupal/redis
+```
+
+* add setting information to `web/sites/default/settings.local.php`; remember to update '<DDEV_SITENAME>'
+
+```php
+# DDEV SERVICE: Redis
```
mradcliffe, line 69: It might be helpful to suggest wrapping this in an if condition `if (getenv('IS_DDEV_PROJECT') == 'true') {}`.

docker-compose-services/redis/README.md
```diff
@@ -49,20 +49,35 @@ ddev redis-cli
...
+$settings['redis.connection']['host']      = 'ddev-${DDEV_SITENAME}-redis';  // Your Redis instance hostname.
```
mradcliffe, line 70: I wasn't able to resolve `${DDEV_SITENAME}` locally even if I loaded settings.local.php after settings.ddev.php in my settings.php, but `$_ENV['DDEV_SITENAME']` (or `getenv('DDEV_SITENAME')` did work. My ddev version is v1.18.2

mradcliffe (changes requested):
I was able to successfully follow the instructions with the exception of the item below:

- I needed to get `DDEV_SITENAME` from `$_ENV['DDEV_SITENAME']`.

Also made a suggestion.

rfay:
How are we with this? 

What I'd really like to do now is pull this but... deprecate it at the same time, and make sure that https://github.com/drud/ddev-redis has everything we want and that everybody is happy with it. It's now officially maintained.

tyler36:
It would be good if `https://github.com/drud/ddev-redis` could automatically add the lines to `settings.ddev.php`

rfay:
Well... nothing should really add to settings.ddev.php, because ddev will then overwrite it. Better settings.local.php or add-on to settings.php. But the current ddev-redis is not Drupal-specific and I doubt that it needs to be?

rfay:
Oh, you were talking about settings.ddev.redis.php - Yes, take a look at try it out please. It does all that thanks to @hussainweb. https://github.com/drud/ddev-redis.  Just try `ddev get drud/ddev-redis`.

I didn't even notice that it was doing all that... It should probably have more info about what happens if it's not a Drupal project.

rfay:
Support for this is moved into the add-on, https://github.com/drud/ddev-redis

PR #195, opened 2022-01-26
use official zookeeper image

mkalkbrenner:
switch from bitnami to official for arm64 support


rfay:
Thanks for doing this! I'm on vacation on a road trip with very limited internet, but hope to look soon.

rfay:
Just a note that after v1.19.0 release this is high priority to get in and also to make into a `ddev get` addon.

rfay (changes requested):
Thanks for all the work on this.

I rebased and made some minor changes. The use of `ddev-<project>-solr` is no longer necessary in ddev v1.19.

1. I wasn't able to figure out what the README instructions mean by `Default Solr collection: techproducts` -there's no entry in the Drupal config form by that name. 
2. I didn't know what to put in "solr core name" as that's not mentioned in the README.
3. zookeeper wouldn't come up, so I got stopped there. See gist at https://gist.github.com/rfay/7238505ac7b7455d50ce3764880dfd11

Thanks for your pioneering of all this. Sorry for the slow review. My hope is to translate this into a `ddev-get` recipe right away, but I have to succeed at setup first.

rfay:
Bitnami now has a multi-arch zookeeper image, https://hub.docker.com/r/bitnami/zookeeper/tags

rfay:
This should be obsolete right? Just add pointer to your new add-on?

Issue #192, opened 2022-01-18
Chrome headless 500 Internal Server Error 


juban:
Hi,

I'm trying to setup a headless chrome as a service following the headless-chrome service (for Behat Testing) and have the following issue:

Service is running fine, and I can access http://localhost:9222 alright from the host.
However, when running `wget http://chrome:9222` from the web container, i constantly get a `500 Internal Server Error` response.
I'm running lastest ddev 1.18.2 and Docker Desktop 4.4.2 on a M1 MacBook Pro.

Does anyone else is experiencing this issue?

rfay:
I assume you're using https://github.com/drud/ddev-contrib/tree/master/docker-compose-services/headless-chrome

It sure sounds like you are running the amd64 version of the chrome image.

Please `docker pull isholgueras/chrome-headless:latest` to make sure you have the right one. 

Then after `ddev restart`, `docker exec -s chrome arch`  and see what it says, amd64 or arm64. 

I'm not sure wget is the right tool to be using there inside the web container though, I'd use `curl -I http://chrome:9222` to start analyzing the connection.

juban:
Thank you @rfay 

Yes, I'm pretty sure I'm using latest image build here (did a pull as suggested anyway to be double sure).

When checking the arch value, i'm getting `aarch64` which is `arm` if I'm not wrong, so should be good.

However, trying to curl the container on port 9222 still gives different results depending on if the request is sent from the host or from within the web container:

From host:
```
curl -I http://localhost:9222
HTTP/1.1 200 OK
Content-Length:1540
Content-Type:text/html; charset=UTF-8
```

From web container:
```
curl -I http://chrome:9222
HTTP/1.1 500 Internal Server Error
Content-Length:63
Content-Type:text/html
```

So, same results using wget, curl or whatever utility allowing to send a web request.

This is odd.

jonaseberle:
The body of that response contains an error message:
`Host header is specified and is not an IP address or localhost.`
wget/curl seem to transmit `Host: chrome:9222` (see `curl -v`) and chrome does not seem to like that.

This does not return an error: 
`curl -v -H 'Host:' chrome http://chrome:9222`

juban:
After double thoughts, the `aarch64` arch value looks strange. The image is available for `arm64` or `amd64`. How come I end up with a `aarch64` version? Could it be related to a docker-compose issue maybe?

juban:
@jonaseberle Thank you for your help.

[...]

PR #196, opened 2022-01-29
Update Puppeteer Chrome recipe to support ddev 1.6 and upper and workaround for Apple M1 arch

juban:
## The New Solution:
This PR provides a reliable way to make Puppeteer run inside the ddev web container (replacing the current "Puppeteer Headless Chrome support" recipe).

The previous recipe was not working anymore on latest ddev web container images (> 1.6) because of outdated packages.
Also, the current chromium bundled version in Puppeteer is not working on Apple Silicon M1 architecture.
This update provides a workaround by overriding the `webimage_extra_packages` configuration to install a native chromium package which puppeteer will use instead of the bundled one.

Follow instructions provided in the recipe `recipes/puppeteer-chromium-support/README.md` file.

## Related Issue Link(s):
https://github.com/drud/ddev-contrib/issues/192 (Not directly related as the "Headless Chrome for Behat Testing" service seams to work fine on M1)


rfay:
Please reach out for some other puppeteer users to review this, thanks!

juban:
@rfay Will do. Do you have anyone in mind I could reach out in the ddev community? @isholgueras or @jonaseberle maybe?

jonaseberle:
Installing headless browsers into the `web` container is not for me. I also don't use Puppeteer but selenium hub with chrome and firefox.

rfay:
Current ddev stable is v1.18, not sure why this is trying to support v1.16 and higher?

juban:
@rfay In fact, the previous recipe stopped working starting ddev 1.16, in other words, the updated recipe will work on ddev 1.16 and upper. Sorry for not being clear.

rfay:
Thanks for the explanation. I'm not sure the difference is worth explaining, or that it's worth encouraging people to use ddev v1.16 :)

juban:
Yes, maybe a simple "latest ddev version" would be better, I agree. I wanted to stick to the contribution guidelines which state that the goal of the update should be specified clearly.
I'll simplified that.

juban:
I fixed the README file accordingly.

rfay:
The best place for ongoing work is https://github.com/drud/ddev-selenium-standalone-chrome

PR #197, opened 2022-02-02
Add drupal pre-commit phpcs recipe running inside DDEV

bserem:
## What is new:
Provide a git-hook recipe that executes PHPCS **from inside DDEV** during `git commit`.

This allows for the host machine to not have PHP and/or PHPCS installed at all, everything is happening inside the container, although the actual commit can happen either inside or outside of the container.

Provides the files and documentation needed.

Follow the instruction in the README, stage a file that would fail in PHPCS and try to commit either from the host machine or from inside the web container in DDEV.

Sample photo (php errors are intentional):
image

## Related Issue Link(s):
https://github.com/drud/ddev/issues/3518 talks about the DDEV verbose line (last line in the above screenshot)

### DIFF
--- a/recipes/git-hooks/pre-commit-drupal-phpcs/pre-commit.php
+++ b/recipes/git-hooks/pre-commit-drupal-phpcs/pre-commit.php
@@ -1,0 +1,119 @@
+<?php
+
+/**
+ * @file
+ * A Git pre-commit hook script to check files for PHP syntax errors and Drupal
+ * coding standards violations. Requires phpcs and Coder Sniffer:
+ *
+ * @see https://drupal.org/node/1419988
+ *
+ * Heavily copied from https://www.drupal.org/project/dcq
+ * and modified for Drupal 8+ and for running **inside** DDEV.
+ *
+ * Ignores all files under config/sync
+ */
+
+const CONFIG_ROOT = 'config/sync';
+
+$return_back = '../../';
+// Building an array of ignored files and folders.
+$file_path = __DIR__ . '/' . $return_back . '.phpcs_ignore';
+$ignore = [];
+$ignore[] = CONFIG_ROOT;
+if (file_exists($file_path)) {
+  $file = fopen($file_path, "r");
+  if ($file) {
+    while (!feof($file)) {
+      $ignore[] = fgets($file);
+    }
+  }
+}
+
+// Extensions of files to test.
+$file_exts = array(
+  'php',
+  'module',
+  'inc',
+  'install',
+  'profile',
+  'test',
+  'theme',
+  'txt',
+  'info',
+  'md',
+  'yml',
+);
+
+$exit_code = 0;
+$files = array();
+$return = 0;
+
+// Determine whether this is the first commit or not. If it is, set $against to
+// the hash of a magical, empty commit to compare to.
+exec('git rev-parse --verify HEAD 2> /dev/null', $files, $return);
+$against = ($return == 0) ? 'HEAD' : '4b825dc642cb6eb9a060e54bf8d69288fbee4904';
+
+// Identify changed files.
+exec("git diff-index --cached --name-only $against", $files);
+
+print "\nPrecommit PHPCS\n\n";
+
+// The number of ignored items.
+$ignored_items_count = count($ignore);
+echo "\033[0;31mYou have $ignored_items_count item(s) in your ignore list!\n\033[0m";
+
+foreach ($files as $file) {
+
+  if (file_exists($file) && !is_dir($file)) {
+
+    // Check files to ignore.
+    if ($ignored_items_count > 0) {
+      foreach ($ignore as $key => $value) {
+        if (substr($file, 0, strlen(trim($value))) == trim($value) && trim($value) != '') {
+          echo "\033[0;31mIgnored in .phpcs_ignore file: $file\n\033[0m";
+          continue 2;
+        }
+      }
+    }
+
+    echo "\033[0;32mChecking file {$file}\033[0m\n\n";
+
+    $ext = pathinfo($file, PATHINFO_EXTENSION);
+
+    if (!in_array($ext, $file_exts)) {
+      continue;
+    }
+
+    $phpcs_output = array();
+
+    $file = escapeshellarg($file);
+    // Add extra path to get warranty for run drush.
+    $dir = str_replace('\'', '', dirname($file));
+    $extra = '';
+    if ($dir !== '.') {
+      $extra = $dir . '/';
+    }
+
+    // Perform PHP syntax check (lint).
+    $return = 0;
+    $lint_cmd = "php -l {$file}";
+    $lint_output = array();
+    exec($lint_cmd, $lint_output, $return);
+    if ($return !== 0) {
+      // Format error messages and set exit code.
+      $exit_code = 1;
+    }
+
+    // Perform phpcs test.
+    $return = 0;
+    $phpcs_cmd = 'cd /var/www/html && phpcs --standard=Drupal --extensions=php,module,inc,install,test,profile,theme ' . $file;
+    exec($phpcs_cmd, $phpcs_output, $return);
+    if ($return !== 0) {
+      // Format error messages and set exit code.
+      echo implode("\n", $phpcs_output), "\n";
+      $exit_code = 1;
+    }
+  }
+}
+
+exit($exit_code);
--- a/recipes/git-hooks/pre-commit-drupal-phpcs/README.md
+++ b/recipes/git-hooks/pre-commit-drupal-phpcs/README.md
@@ -1,0 +1,19 @@
+# Drupal pre-commit PHPCS
+
+Runs PHPCS on the files to be commited from **inside** DDEV, regardless if
+you are committing from your host machine or from inside the container.
+
+## Installation
+
+Copy `pre-commit` and `pre-commit.php` under `PROJECT_ROOT/scripts/git/`
+and then:
+```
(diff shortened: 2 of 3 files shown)


rfay (changes requested):
Please make sure to
* Add to the top-level readme.md
* Add a footer of your readme saying "contributed by" like the others have
* Mention if you haven't that you can't do this unless you have php installed on the host (many people don't because they just use it inside ddev's web container)

I haven't reviewed carefully, just noted these things. The hook looks pretty complicated. I also recommend a pre-push hook instead of pre-commit.

recipes/git-hooks/pre-commit-drupal-phpcs/pre-commit.php
```diff
@@ -0,0 +1,121 @@
...
+    $lint_output = [];
+    exec($lint_cmd, $lint_output, $return);
+    if ($return !== 0) {
+      // Format error messages and set exit code.
+      $exit_code = 1;
+    }
+
+    // Perform phpcs test.
+    $return = 0;
+    $phpcs_cmd = 'cd /var/www/html && phpcs --standard=Drupal --extensions=php,module,inc,install,test,profile,theme ' . $file;
```
tyler36, line 111: Could this be modified to use a local config instead?

This would allow this PR to support other projects types (Eg. Laravel, WP).

https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#using-a-default-configuration-file

I'm not sure if the command switch take precedent over config but hardcoding the standard & extensions limits long-term flexability & usability.

bserem:
@rfay this recipe allows phpcs to run **without php installed on the host machine**. I made that clearer on the readme.
At the moment I do not have the time to change this to a `pre-push` hook, but I really like the idea and will try to do this soon.

@tyler36 I do not have any experience with other projects, this is what Drupal recommends for custom/contrib code. Since this is a recipe, anybody can freely change it to their needs. I am positive it can be altered as one desires.

The whole idea behind it is to remove the necessity for PHP on the host machine, so that our developers only need to care about their code.

_PS: A follow-up for `commit-msg` is coming within the day_

rfay:
This also requires phpcs to be required where you want to use it, either host or in ddev web container, right?

tyler36:
@bserem Yes. 

But if you changed the command line

[recipes/git-hooks/pre-commit-drupal-phpcs/pre-commit.php](https://github.com/drud/ddev-contrib/pull/197/files/82ce2c00f9c9d84fa1b610bcfca5299579ad1f2e#diff-b53806955634a8c1ba38fc603e54a663c6456bd8dfb9f28a3ac4dc2e4187aa4a)

```diff
- $phpcs_cmd = 'cd /var/www/html && phpcs --standard=Drupal --extensions=php,module,inc,install,test,profile,theme ' . $file;
+ $phpcs_cmd = 'cd /var/www/html && phpcs ' . $file;
```

- and include a `phpcs.xml`

```xml
<?xml version="1.0" encoding="UTF-8"?>
<ruleset name="my_drupal_module">
    <description>PHP CodeSniffer configuration for a Drupal module.</description>

    <arg name="extensions" value="php,inc,module,install,info,test,profile,theme,css,js"/>
    <arg name="report" value="full"/>
    <arg value="p"/>
    <file>.</file>

    <!-- Exclude third party libraries. -->
    <exclude-pattern>./vendor</exclude-pattern>

    <rule ref="Drupal" />

</ruleset>
```

This would do exactly the same thing. However, it decouples the solution from specific Drupal-only config, to a extenadable pre-commit PHPCS reciepe.

bserem:
@rfay I improved the docs to add the dependencies.

@tyler36 I am also ignoring drupal generated yml files in the start of the script, and if I don't do that then the hook takes a long time to execute. In the end the `xml` file will ignore them, but it will loop through all of them because of the way the hook is structured (to check only the files to be committed).

Switching to a `phpcs.xml` file is possible like you demonstrated, and it would need the script to be adapted like the following, **but** it would take a lot of time to run because the ignore doesn't happen on php level.
```php
$return_back = '../../';
// Building an array of ignored files and folders.
$file_path = __DIR__ . '/' . $return_back . '.phpcs_ignore';
$ignore = [];
$ignore[] = CONFIG_ROOT;
if (file_exists($file_path)) {
  $file = fopen($file_path, "r");
  if ($file) {
    while (!feof($file)) {
      $ignore[] = fgets($file);
    }
  }
}

$exit_code = 0;
$files = [];
$return = 0;

// Determine whether this is the first commit or not. If it is, set $against to
// the hash of a magical, empty commit to compare to.
exec('git rev-parse --verify HEAD 2> /dev/null', $files, $return);
$against = ($return == 0) ? 'HEAD' : '4b825dc642cb6eb9a060e54bf8d69288fbee4904';

// Identify changed files.
exec("git diff-index --cached --name-only $against", $files);

print "\nPrecommit PHPCS\n\n";

// The number of ignored items (including CONFIG_ROOT).
$ignored_items_count = count($ignore);
echo "\033[0;31mYou have " . ($ignored_items_count - 1) . " item(s) in your ignore list!\n\033[0m";

foreach ($files as $file) {

  if (file_exists($file) && !is_dir($file)) {

    // Check files to ignore.
    if ($ignored_items_count > 0) {
      foreach ($ignore as $key => $value) {
        // Display a message for ignored files.
        if (substr($file, 0, strlen(trim($value))) == trim($value) && trim($value) != '') {
          echo "\033[0;31mIgnored in .phpcs_ignore file: $file\n\033[0m";
          continue 2;
        }
      }
    }

    echo "\033[0;32mChecking file {$file}\033[0m\n\n";

    $ext = pathinfo($file, PATHINFO_EXTENSION);

    $phpcs_output = [];

    $file = escapeshellarg($file);
    // Add extra path to get warranty for run drush.
    $dir = str_replace('\'', '', dirname($file));
    $extra = '';
    if ($dir !== '.') {
      $extra = $dir . '/';
    }

    // Perform PHP syntax check (lint).
    $return = 0;
    $lint_cmd = "php -l {$file}";
    $lint_output = [];
    exec($lint_cmd, $lint_output, $return);
    if ($return !== 0) {
      // Format error messages and set exit code.
      $exit_code = 1;
    }

    // Perform phpcs test.
    $return = 0;
    $phpcs_cmd = 'cd /var/www/html && phpcs ' . $file;
    exec($phpcs_cmd, $phpcs_output, $return);
    if ($return !== 0) {
      // Format error messages and set exit code.
      echo implode("\n", $phpcs_output), "\n";
      $exit_code = 1;
    }
  }
}

exit($exit_code);
```

I added an entry to the documentation towards your code and suggestions, so that people interested can find it easily.

rfay:
Thanks so much, this will be a nice contribution. I'm on vacation for a couple more weeks, so may be some time before I get to review it, but if somebody else manually tests/reviews I'll pull it.

bserem:
Enjoy the vacation!

bserem:
@tyler36 after reading a bit about the `phpcs.xml` file and giving it a spin for a week to make sure I got things right, I updated the PR and the script per your suggestions. Now all configuration happens inside the `phpcs.xml` file so the script is more readable and does not need to be altered to be used to another project.

Please let me know what you thing about it in the current state and do not hesitate to suggest things.

Side bonus: the pre-commit script and the project phpcs configuration are now the same.

rfay (commented):
Thanks - The things I know to check are fine. If @tyler36 or anybody else can manually test and review this I'll pull it. If you know somebody that's using it or could test, just send them over. Thanks!

tyler36:
Great job!

- Works well on a WSL host.
- Didn't work on Win10 host (maybe add a note in the docs?)
- VSCode extension valeryanm.vscode-phpsab can use the same PHPCS file (yeah!)

recipes/git-hooks/pre-commit-phpcs/pre-commit
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# Run pre-commit check PHP script inside ddev when committing from host.
+if [ "$IS_DDEV_PROJECT" != true ]; then
+  ddev exec /usr/bin/php scripts/git/pre-commit-phpcs.php
+else
+  /usr/bin/php scripts/git/pre-commit-phpcs.php
```
tyler36, line 7: Is `php` always run from this location on a host? 
Maybe change to below so it can be picked up from $PATH

```
  php scripts/git/pre-commit-phpcs.php
```

rfay: On most systems it will be in PATH. Same goes for `ddev exec`, it can just be `ddev exec php`

tyler36: I was concerned about `else` version more so because of the variety of hosts (Win, WSL, Linux, MacOS).

bserem: @tyler36 the `else` part means you are inside the container, so it will be always a ddev container (Linux) and `/usr/bin/php` is the right path.

Remember, this script does not use host php, it detects whether you are committing from the host or from the container.

tyler36: You're right. 

Sorry for the confusion. Still waiting for the coffee to kick in.

bserem: No worries. I'll resolve the conversation then.

Please let Randy know if this is ok now so that it gets merged. An extra check to the documentation would be great if you have the time. I was careful, but you never know.

tyler36 (approved):

PR #208, opened 2022-03-16
Rid extra characters from docker-compose.chrome

makraz:

Rid extra characters from docker-compose.chrome.yaml

Running docker composer without issue


rfay (approved):
Hmm, wonder what those were doing there!

PR #213, opened 2022-06-09
Bump meilisearch to 0.27.2, adds arm64 support

drifteaur:

The previous meilisearch image did not support arm64/Apple Silicon

The image is upgraded to version 0.27.2 (there's arm64 support since version 0.25). The documentation is updated accordingly.

Use the provided `docker-compose.meilisearch.yaml` file. After starting, the service should be available at: `http://ddev-<projectname>-meilisearch:7700`

## Related Issue Link(s):

#186


rfay (approved):
Awesome, thanks! You should consider making a ddev-get add-on recipe!

PR #214, opened 2022-07-05
Add php8_2 as a copy of php8_1 with version changes

Berdir:
Tested locally, have a "working" D10 project on PHP 8.2.


rfay (approved):
Thanks!

rfay:
It would be even better to turn this into a ddev-get add-on, which would allow people to do it with one command. It's really easy to set up.  https://github.com/drud/ddev-addon-template

PR #216, opened 2022-08-04
Sqlsrv fixes

NickStees:
SQLSRV documentation fixes


NickStees:
Note: I used `php7.3-dev` but it looks like what I did in the PR `php-dev` should work for newer systems. But I haven't tested that I should...  see:https://stackoverflow.com/questions/3108937/how-to-install-and-run-phpize

NickStees:
So I tried to get that docker file working with PHP 8.0 but kept getting the phpize errors. The only way I could get that SQLSRV container running is with PHP 7.4 so I guess ill update the example file to 7.4?

rfay:
You'll want `php8.0-dev` for php 8.0.

PR #220, opened 2023-02-03
🚚 move to `ddev` domain

tyler36:
This PR updates docs for move to `ddev` domain.

### DIFF
--- a/recipes/proxy/Dockerfile
+++ b/recipes/proxy/Dockerfile
@@ -1,7 +1,7 @@
 # This is a Dockerfile for use with the ddev-webserver image
 # It should be in .ddev/web-build/Dockerfile
 # and then you can expand on it.
-ARG BASE_IMAGE=drud/ddev-webserver:v1.9.1
+ARG BASE_IMAGE=ddev/ddev-webserver:v1.9.1
 FROM $BASE_IMAGE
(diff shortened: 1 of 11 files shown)


rfay:
There are no tests in ddev-contrib. That's the great thing about the move to add-ons of course.

recipes/proxy/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # This is a Dockerfile for use with the ddev-webserver image
 # It should be in .ddev/web-build/Dockerfile
 # and then you can expand on it.
-ARG BASE_IMAGE=drud/ddev-webserver:v1.9.1
+ARG BASE_IMAGE=ddev/ddev-webserver:v1.9.1
```
rfay, line 4: This isn't correct, the dockerhub repo will not yet be updated to use drud. Hopefully we can do that later. Not sure we have control of it.

rfay (changes requested):
It's hard to do things that refer to ddev/ddev at this point, as it hasn't moved. And things that refer to dockerhub can't be changed at this point.

recipes/proxy/Dockerfile
tyler36: reverted.

rfay (approved):
Thanks!

Issue #223, opened 2023-03-22
SQL Server (Microsoft) install sqlsrv error


robertoperuzzo:
Using the following Dockerfile in my project 

```
ARG BASE_IMAGE
FROM $BASE_IMAGE

ENV PATH="${PATH}:/opt/mssql-tools/bin"
RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests apt-utils curl gnupg2 ca-certificates

RUN curl -sSL -O https://packages.microsoft.com/keys/microsoft.asc
RUN apt-key add <microsoft.asc
RUN curl -sSL -o /etc/apt/sources.list.d/mssql-release.list https://packages.microsoft.com/config/debian/11/prod.list

RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests build-essential dialog php-pear php8.1-dev unixodbc-dev locales

RUN ACCEPT_EULA=Y DEBIAN_FRONTEND=noninteractive apt-get install -y msodbcsql18 mssql-tools

# Change the PHP version to what you want. It is currently set to version 8.0.
RUN pecl channel-update pecl.php.net
RUN pecl -d php_suffix=8.1 install sqlsrv
RUN pecl -d php_suffix=8.1 install pdo_sqlsrv

RUN echo 'extension=sqlsrv.so' >> "/etc/php/8.1/mods-available/sqlsrv.ini"
RUN echo 'extension=pdo_sqlsrv.so' >> "/etc/php/8.1/mods-available/pdo_sqlsrv.ini"

RUN phpenmod -v 8.1 sqlsrv pdo_sqlsrv

# Add AWS CLI.
RUN wget https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip -O awscliv2.zip
RUN unzip awscliv2.zip
RUN ./aws/install
```
I'm getting the error 
```
❯ ddev debug refresh
failed to solve: executor failed running [/bin/sh -c pecl -d php_suffix=8.1 install sqlsrv]: exit code: 1
...
...
...
#22 18.22 /bin/sed: can't read /usr/lib/x86_64-linux-gnu/libltdl.la: No such file or directory
#22 18.22 libtool:   error: '/usr/lib/x86_64-linux-gnu/libltdl.la' is not a valid libtool archive
[...]

PR #224, opened 2023-03-22
feat ref #223: fixes the error installing sqlsrv

robertoperuzzo:
* uses odbc version 2.3.7
* parameterizes PHP version

Using the previous Dockerfile I get the following error
```
❯ ddev debug refresh
failed to solve: executor failed running [/bin/sh -c pecl -d php_suffix=8.1 install sqlsrv]: exit code: 1
...
...
...
#22 18.22 /bin/sed: can't read /usr/lib/x86_64-linux-gnu/libltdl.la: No such file or directory
#22 18.22 libtool:   error: '/usr/lib/x86_64-linux-gnu/libltdl.la' is not a valid libtool archive
#22 18.22 make: *** [Makefile:250: sqlsrv.la] Error 1
#22 18.23 ERROR: `make' failed
...
...
```
It seems something related to version 2.3.11 of unixodbc.

Forcing the unixodbc version 2.3.7

1. Copy the `Dockerfile` in `.ddev/web-build/`
2. run `ddev php --info | grep pdo_sqlsrv`

You should get 
```
/etc/php/8.1/cli/conf.d/20-pdo_sqlsrv.ini,
pdo_sqlsrv
pdo_sqlsrv support => enabled
pdo_sqlsrv.client_buffer_max_kb_size => 10240 => 10240
pdo_sqlsrv.log_severity => 0 => 0
pdo_sqlsrv.report_additional_errors => 1 => 1
pdo_sqlsrv.set_locale_info => 2 => 2
```

### DIFF
--- a/docker-compose-services/sqlsrv/Dockerfile
+++ b/docker-compose-services/sqlsrv/Dockerfile
@@ -1,5 +1,4 @@
-ARG BASE_IMAGE
-FROM $BASE_IMAGE
+ARG odbc_version=2.3.7
 
 ENV PATH="${PATH}:/opt/mssql-tools/bin"
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests apt-utils curl gnupg2 ca-certificates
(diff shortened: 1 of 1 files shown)


docker-compose-services/sqlsrv/Dockerfile
```diff
@@ -1,5 +1,4 @@
-ARG BASE_IMAGE
-FROM $BASE_IMAGE
+ARG odbc_version=2.3.7
```
rfay, line 1: Shouldn't this be ENV? I don't see where you're using it as an arg. Here you're using it as an argument with a default value.

rfay (changes requested):
@drupal-daffie please review and manually test this and see if you approve. Assuming you're still familiar with this setup.

docker-compose-services/sqlsrv/Dockerfile
robertoperuzzo: I used ARG instead of ENV because I need this variable in image building only (not in the container environment). I use it in rows from 15 to 18 

```
...
    unixodbc=$odbc_version \
    odbcinst=$odbc_version \
    odbcinst1debian2=$odbc_version \
    unixodbc-dev=$odbc_version \
...
```

rfay: I can go with that :)

rfay:
I haven't manually tested, but hope @drupal-daffie will take a look

rfay:
Actually, what I'd love to have you do is turn this into a ddev-get add-on and then maintain it, so we wouldn't find out things like this by accident. https://ddev.readthedocs.io/en/latest/users/extend/additional-services/#creating-an-additional-service-for-ddev-get

Would you consider doing that?

robertoperuzzo:
> Actually, what I'd love to have you do is turn this into a ddev-get add-on and then maintain it, so we wouldn't find out things like this by accident. https://ddev.readthedocs.io/en/latest/users/extend/additional-services/#creating-an-additional-service-for-ddev-get
> 
> Would you consider doing that?

It sounds good. I'm going to read the add-on documentation and try do create the new add-on.

robertoperuzzo:
@rfay I created this repo for the ddev add-on https://github.com/robertoperuzzo/ddev-sqlsrv

robertoperuzzo:
> @rfay I created this repo for the ddev add-on https://github.com/robertoperuzzo/ddev-sqlsrv

I have to fix the README and the tests.

rfay:
Awesome!

You'll probably want to change this PR to a "Go to the maintained add-on instead" notice.

rfay:
Let's get this in and please just add a reference to your *maintained* add-on. Thanks so much!

robertoperuzzo:
> Let's get this in and please just add a reference to your _maintained_ add-on. Thanks so much!

I'm sorry but I didn't understand what _"add a reference to your _maintained_ add-on"_ means.

rfay:
Just update ddev-contrib here to point to ddev/ddev-sqlsrv and point out that the existing recipe here isn't the authoritative one any more.

drupal-daffie:
I have not test the code, but all the changes look good to me.

PR #225, opened 2023-04-05
Add Silverstripe sake

Firesphere:

Add Silverstripe sake to remove the need to fully type out `ddev exec vendor/bin/sake {task}`

With this addition, one can easily install the sake proxy, and run `ddev sake {task}`

Run the commands described above.

### DIFF
--- a/custom-commands/silverstripe/sake
+++ b/custom-commands/silverstripe/sake
@@ -1,3 +1,9 @@
 #!/bin/bash
 
-vendor/bin/sake $@
+## Description: Run Silverstripe make
+## Usage: sake <command-url> <parameters>
+## Example: "ddev sake dev/tasks" or "ddev sake dev/build flush=all"
+## ExecRaw: true
+## HostWorkingDir: true
+
+sake $@
--- a/custom-commands/silverstripe/sake
+++ b/custom-commands/silverstripe/sake
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+vendor/bin/sake $@
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ DDEV's [custom commands](https://ddev.readthedocs.io/en/latest/users/extend/cust
 * [Stop all running projects except the current](custom-commands/stop-other)
 * [Dynamically enable / disable a service](custom-commands/dynamic-service)
 * [Automatically open browser and login to Drupal](custom-commands/drupal-login)
+* [Run Silverstripe frameworks command directly](custom-commands/silverstripe)
 
 ## Additional services added via docker-compose.\<service\>.yaml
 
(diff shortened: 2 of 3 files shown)


rfay (approved):
I'm sure this is fine, but you might consider creating a DDEV add-on with more facilities than this (and more usable and discoverable). https://ddev.readthedocs.io/en/latest/users/extend/additional-services/

rfay (changes requested):
Oh, you need to add this to the top-level readme or nobody will ever see it. Just add a commit to do that.

Firesphere:
> I'm sure this is fine, but you might consider creating a DDEV add-on with more facilities than this (and more usable and discoverable). https://ddev.readthedocs.io/en/latest/users/extend/additional-services/

That was indeed a possible next step for me, simply haven't had time, and made this in the interim for my colleagues. Figured might as well PR at :)

Firesphere:
Just added to the main readme.

I'll have a look at making it a proper additional service later, but for now, a useful start for Silverstripe users, I think

custom-commands/silverstripe/sake
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+vendor/bin/sake $@
```
rfay, line 3: Sorry, I should have said this before, but `vendor/bin` is already in the $PATH, so you could just have `sake` here.

custom-commands/silverstripe/sake
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
rfay, line 1: This should really have the normal annotations on it, https://ddev.readthedocs.io/en/latest/users/extend/custom-commands/#annotations-supported - at least add the name and such.

You may especially want `HostWorkingDir`.  It works for example with `ddev yarn`, executes in the relative directory with the root. You can look at the yarn command for an example.

rfay (changes requested):
Sorry I didn't look closely before, but think you'll want to take a look at these things.

custom-commands/silverstripe/sake
Firesphere: Oh, cool! That indeed works just fine! :+1:

custom-commands/silverstripe/sake
Firesphere: Yes, done, thank you!

Firesphere:
> Also, a PR would be welcome to the quickstarts

I've written a sort-of quick start for Silverstripe here:
https://firesphere.dev/articles/ddevelopment-environment/#e44

I'll see if I can formalize that in to a more useful/complete documentation for the quickstarts!

rfay:
Cool - the only problem I saw on a quick scan was I always recommend using `ddev composer` rather than composer on the host, because composer on the host may not match the php configuration (or composer version) of your actual setup inside container.

PR #229, opened 2023-07-20
Bump meilisearch to 1.2.0, update README

danielkestler:

Update meilisearch version to 1.2.0.


rfay:
This is fine with me, but would you please create a DDEV add-on instead? It will have a much better life, and you'll be able to actually maintain it. Things are just lost here in ddev-contrib. 

* https://github.com/ddev/ddev-addon-template
* https://ddev.readthedocs.io/en/latest/users/extend/additional-services/

If you have carefully manually tested this, I'll pull it, but please do create an add-on with tests.

danielkestler:
I'm not the author of the original ddev-contrib part for meilisearch, I just bumped the version number here. Will have a look at it.

rfay:
Would love to have you make an add-on. You'd give credit to the original author in the README of course. Let me know when you've tested this manually and I'll pull it here.

PR #233, opened 2023-08-04
Add-ons on top

hanoii:
I think add-ons should be actually more prominent now, thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # ddev-contrib: Contrib for DDEV
 
-Contrib space for DDEV services, tools, snippets, and approaches.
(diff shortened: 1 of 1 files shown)


rfay:
Can we go a step beyond and say that these things are typically handled with add-ons, both official and third-party, and link to the docs? There's an incomplete list of official ones in the docs, and of course the current `ddev get --list --all`, a reference to that would be great. 

I know it's more than you signed up for but it's what's needed here, to say that this repo is mostly obsolete.

rfay:
I noticed we lost this, can you follow up and rebase and I'll pull?

PR #235, opened 2023-09-24
Create CiviCRM recipe, fixes ddev/ddev#5373

gitressa:

Here's a MR for creating a CiviCRM recipe, as discussed in https://github.com/ddev/ddev/issues/5373.

Adds documentation for CiviCRM.


rfay (changes requested):
Thanks for this! 
* Please add an entry pointing to it in the top-level readme, or nobody will know it's here.
* Please add a `Contributed by` at the bottom so people could at least mention you in a followup.

gitressa:
You're welcome! I added link and contributed by.

Do you have an opinion on my two questions in https://github.com/ddev/ddev/issues/5373#issuecomment-1732555587?

rfay (approved):
Thanks!

PR #236, opened 2023-09-25
Remove ext-intl PHP note, since it is included

gitressa:

The `ext-intl` PHP extension is included in DDEV, so removing this note.


gitressa:
Thanks!

PR #240, opened 2023-12-05
Make PHP Version Dynamic

wolffc:
Use $DDEV_PHP_VERSION to update the correct version.

instead of hardcoded php version number use $DDEV_PHP_VERSION variable to make this dynamic.

that there is no fixed 7.4 in the files so  the solution is better managable if projects evolve. 

`ddev up`  wait for the grpc / protobuf build to complete and check that the extensions are installed `ddev exec php -m | grep -i 'grpc\|protobuf'`

### DIFF
--- a/web-container-dockerfiles/grpc/Dockerfile
+++ b/web-container-dockerfiles/grpc/Dockerfile
@@ -1,12 +1,12 @@
 ARG BASE_IMAGE
 FROM $BASE_IMAGE
 
-ENV PHP_VERSION=7.4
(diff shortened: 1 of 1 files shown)


tyler36:
Thanks for the PR.

Side note: This site is slowly dying. If this is something you use a lot, consider creating a DDEV addon. 
The Discord community is always happy to help if you need a hand. https://github.com/ddev/ddev-addon-template 😄

rfay:
I'll pull this @wolffc but if you could please update the top-level readme (which is 404) it would be good too. But as @tyler36 says, we don't really use this any more.

PR #242, opened 2024-01-19
Properly handle special characters and update output (dump-and-deploy)

jonaseberle:
This updates the dump-and-deploy custom command. It had some problems that were fixed over the time.

    Properly handle special characters and update output (dump-and-deploy)
    
    This deals with special characters in passwords (and everywhere else).
    
    It lints code and improves the output.

After following the installation instructions the simplest to test is to run `ddev dump-db ddev` on a project database and afterwards `ddev deploy-db ddev` (the only side-effect is that it creates the directory `./data/sql`).

### DIFF
--- a/custom-commands/dump-and-deploy-db/README.md
+++ b/custom-commands/dump-and-deploy-db/README.md
@@ -14,7 +14,7 @@ Copy [the web/ directory](web/) into your project's .ddev/commands/web.
 
 ## Configuration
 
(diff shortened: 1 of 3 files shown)


rfay (approved):
I'm sure this is correct because you're the expert. It sure would be better as a DDEV add-on though, where it can have a repo and issues and etc.

PR #244, opened 2024-03-25
Updated Ports - Changed from ports to expose (docker-compose.chrome.yaml)

XjSv:
Changed 'ports' to 'expose'. Ports exposed to the other services but not to the host machine.

That approach usually isn’t sustainable because two projects might want to use the same port, so we expose the additional port to the Docker network and then use ddev-router to bind it to the host. This works only for services with an HTTP API, but results in having both HTTP and HTTPS ports (9998 and 9999).

https://ddev.readthedocs.io/en/latest/users/extend/custom-compose-files/#docker-composeyaml-examples

### DIFF
--- a/docker-compose-services/headless-chrome/docker-compose.chrome.yaml
+++ b/docker-compose-services/headless-chrome/docker-compose.chrome.yaml
@@ -6,7 +6,7 @@ services:
         container_name: ddev-${DDEV_SITENAME}-chrome
         labels:
             com.ddev.site-name: ${DDEV_SITENAME}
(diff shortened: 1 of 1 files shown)


rfay (approved):
This is correct... but I didn't think anybody ever used this any more. There are so many good add-ons that do it better. For example, https://github.com/ddev/ddev-selenium-standalone-chrome

I guess it does say that in the [README](https://github.com/ddev/ddev-contrib/tree/master/docker-compose-services/headless-chrome)

PR #246, opened 2024-06-08
Update chromedriver service and instructions to match latest Drupal 11 latest

scotteuser:
Thanks for this very helpful chromedriver docker compose and detailed instructions, has been very helpful as a reference!

- The default sample chromedriver docker compose file has some issues:
  - The mysql connection sample does not work as is
  - chromeOptions is deprecated and should be goog:chromeOptions
  - adding version is deprecated
- The README is missing some instructions
  - It specifies only leaving the mink args unchanged, but actually the sample docker compose provides more values than that

- Updates docker compose file to fix the above
- Updates the README to fix the above

### DIFF
--- a/docker-compose-services/drupalci-chromedriver/README.md
+++ b/docker-compose-services/drupalci-chromedriver/README.md
@@ -14,21 +14,21 @@ It is based on [Matt Glaman's excelent blog post](https://glamanate.com/blog/run
 
 * Copy `docker-compose.chromedriver.yaml` to the `.ddev` folder of your project.
 * Set environment variables in the "environment" section of the "web" service in `docker-compose.chromedriver.yaml` as appropriate. The defaults are for Drupal Functional and FunctionalJavascript testing.
-* Verify that your phpunit.xml does not define `MINK_DRIVER_ARGS_WEBDRIVER`. This should come from our docker-compose.chromedriver.yaml.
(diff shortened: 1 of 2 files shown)


scotteuser:
I think this is for @heddn and @mglaman as the contributors of this portion of the ddev contrib as per the readme list of contributors https://github.com/ddev/ddev-contrib/blob/master/docker-compose-services/drupalci-chromedriver/README.md

Thank you!

rfay:
I don't mind pulling this, but I don't think anybody uses this any more. Instead, they use the well-maintained https://github.com/ddev/ddev-selenium-standalone-chrome - see instructions there.

scotteuser:
Oh I see! Should I update the README to make that more clear? At the moment it seems to indicate it just as an alternative rather than the preferred way, and though it appeared before the Apple M1 comment I thought it was intended to be for those on such devices.

Or could be removed altogether/marked as deprecated? If it is sticking around makes sense to keep it up to date with latest Drupal I guess though (as is it doesn't work as per the issue description).

Thanks!
Scott

rfay:
Yes, please update the README. I think everybody in the Drupal world is using that add-on. And probably we should be saying "Apple Silicon" instead of "M1" these days. You can update this with another commit or do another PR. Thanks for taking care of this! Mostly ddev-contrib has been replaced with better-maintained add-ons.

rfay (approved):
Thanks!

scotteuser:
Thanks!

PR #247, opened 2024-07-17
Mention paths for TYPO3 v12

linawolf:

Path have changed with TYPO3 v12

Update the docs


rfay (commented):
Isn't this completely replaced by the well-maintained https://github.com/ddev/ddev-typo3-solr ? I would think it would be better to put a big banner at the top saying "use ddev-typo3-solr' ?

However, I'm happy to pull this if you want it. But would prefer for people to use maintained resources.

linawolf:
Thanks a lot, yes that is much better and enabled me to use solr with TYPO3 in ddev. I think the info box is a good idea

rfay:
I'll pull this if you'll follow up with a fix with a LOUD pointer to point people to the add-on :)

PR #248, opened 2024-10-03
Update docker-compose.solr.yaml

thegbomb:
Adding changes to supress warnings.

When using acquia_search setting things up locally takes a bit of work. I have been using this contrib project with the following tweaks.

This stops warings about version being out of date and allows ddev to connect to the solr container from within Drupal.

## Caveat:
Now that Acquia have updated their Solr servers this workaround may no longer be needed and so the caveat in the README.md file at: https://github.com/ddev/ddev-drupal-solr may need to be updated to say that the project can now be used with acquia_search

### DIFF
--- a/docker-compose-services/solr-7/docker-compose.solr.yaml
+++ b/docker-compose-services/solr-7/docker-compose.solr.yaml
@@ -21,8 +21,6 @@
 #   accessed at the URL: http://solr:8983/solr/dev (inside web container)
 #   or at http://myproject.ddev.site:8983/solr/dev (on the host)
 
-version: '3.6'
-
 services:
   solr:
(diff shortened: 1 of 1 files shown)


thegbomb:
Thanks,

I wasn't expecting this to be merged TBH, due to the fact that it was abandoned. It was more to ensure I didn't lose the tweaks, if I ever need them again. I'll look into the https://github.com/ddev/ddev-solr to see if that works.

I was going to do anything to the fading https://github.com/ddev/ddev-drupal-solr, but mentioned it as it indirectly says to use Solr 7 to connect to Acquia and that usually leads to looking here.

All good now, thank you

